PR #1, opened 2019-11-17
Support interrupt conditions

AlexanderPrendota:


AlexanderPrendota:
@wayfarer-rus Let's do improvement in current branch  =)

wayfarer-rus:
removed all unnecessary entities. 

code looks much cleaner like that.

we lost ability to accumulate output buffers if timeout happens though.

src/main/kotlin/com/compiler/server/executor/JavaExecutor.kt
```diff
@@ -2,25 +2,34 @@ package com.compiler.server.executor
...
+    const val MAX_OUTPUT_SIZE = 100 * 1024
+    const val EXECUTION_TIMEOUT = 10000L
+  }
+
   data class ProgramOutput(
     val standardOutput: String,
     val errorOutput: String,
     val exception: Exception? = null
   ) {
+    constructor(exception: Exception) : this("", "", exception)
```
AlexanderPrendota: Why u need the constructor? We can use default values instead

src/main/kotlin/com/compiler/server/executor/JavaExecutor.kt
```diff
@@ -2,25 +2,34 @@ package com.compiler.server.executor
...
+
     fun asExecutionResult(): JavaExecutionResult {
       return JavaExecutionResult(
-        text = "<outStream>$standardOutput\n</outStream>",
+        text = "<outStream>$standardOutput\n</outStream><errStream>$errorOutput</errStream>",
```
AlexanderPrendota: Hm? error after standard?
```
System.out("1")
System.err("123")
System.out("2")
```
probably the output will be
1
2
123

does it correct?)

src/main/kotlin/com/compiler/server/executor/JavaExecutor.kt
```diff
@@ -29,72 +38,94 @@ class JavaExecutor {
 
   fun execute(args: List<String>): JavaExecutionResult {
     return Runtime.getRuntime().exec(args.toTypedArray()).use {
+      // why?
```
AlexanderPrendota: Drop why

src/main/kotlin/com/compiler/server/executor/JavaExecutor.kt
wayfarer-rus: No, it is simpler.

If you have tag <outStream>, why not to have also <errStream>?

You should think about using it somehow, though

src/main/kotlin/com/compiler/server/executor/JavaExecutor.kt
wayfarer-rus: Good idea!

I write too much Java code lately.

PR #3, opened 2019-12-13
New Java executor

AlexanderPrendota:


src/main/kotlin/com/compiler/server/model/ProgramOutput.kt
```diff
@@ -1,27 +1,51 @@
...
-      exception = exception?.let {
-        ExceptionDescriptor(it.message ?: "no message", it::class.java.toString())
-      })
+  fun asExecutionResult(): ExecutionResult {
+    return when {
+      restriction != null -> ExecutionResult().apply { text = buildRestriction(restriction) }
+      exception != null -> ExecutionResult().also { it.exception = exception.toExceptionDescriptor() }
+      else -> {
+        // coroutines can produced incorrect output. see example in `base coroutines test 7`
+        if (standardOutput.startsWith("{")) outputMapper.readValue(standardOutput, ExecutionResult::class.java)
```
AlexanderPrendota: @wayfarer-rus rly? even over here?

src/main/kotlin/com/compiler/server/model/ExceptionDescriptor.kt
```diff
@@ -10,3 +11,8 @@ open class ExceptionDescriptor(
   val cause: ExceptionDescriptor? = null,
   val localizedMessage: String? = null
 )
+
+fun Throwable.toExceptionDescriptor(): ExceptionDescriptor {
+  val rawException = outputMapper.writeValueAsString(this)
```
AlexanderPrendota: even here? )

src/main/kotlin/com/compiler/server/model/ProgramOutput.kt
wayfarer-rus: Yep "Value over the beauty".
I can't see, how you can make it better right now.

PR #5, opened 2020-01-02
Use kotlin version for lib directory name

AlexanderPrendota:
have to prevent local conflicts after updating kotlin version


build.gradle.kts
```diff
@@ -1,24 +1,21 @@
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
-object BuildProps {
-    const val version = "1.3.60-release-155"
-}
+val kotlinVersion: String by System.getProperties()
```
wayfarer-rus: what is this witchcraft? could you give a comment why it will resolve to the correct 'kotlinVersion'?

src/main/resources/libraries.properties
```diff
@@ -0,0 +1,3 @@
+kotlin.version=1.3.60-release-155
+libraries.folder.jvm=1.3.60-release-155
+libraries.folder.js=1.3.60-release-155-js
```
wayfarer-rus: IMO, this file should be removed from git, while it is autogenerated, isn't it?

AlexanderPrendota: won't generate if u just run the application

build.gradle.kts
AlexanderPrendota: I get kotlinVersion from gradle.properties

```
systemProp.kotlinVersion=1.3.60-release-155 
```

src/main/resources/libraries.properties
AlexanderPrendota: just make more understanding for external contributors

wayfarer-rus: write a comment-line at the beginning of the file. Like,
```
# this file is autogenerated by build.gradle.kts
```

AlexanderPrendota: heh, make sense, thanks!

Issue #33, opened 2020-01-22
Webpack bundle support


rjaros:
Any chance of supporting Kotlin/JS features like bundling code with webpack?

AlexanderPrendota:
Hello! Thanks for the feedback. Could you please add a little bit more description of that feature? It's a feature for Kotlin JS team or for this repository?

Thanks.

rjaros:
As far as I understand webpack integration is part of the new Kotlin/JS gradle plugin. I'm not quite sure how it relates to the Kotlin/JS compiler or to this project.

AlexanderPrendota:
So, that is a good feature request. But Actually, I don't have plans to implement that feature but you could create an issue in [Kotlin](https://kotl.in/issue) and maybe someone could help you.🙂

rjaros:
I think I'll try to play with your project, because I think it could be done with not much effort. I'll make a PR if I succeed.

AlexanderPrendota:
Thanks. I will be waiting for your PR 😊

rjaros:
I've opened a PR (#37). I've tested it successfully with my KVision framework (with a lot of npm dependencies from framework jar's). I would like to add Kotlin DCE support as well, but currently I have no idea if it's possible without Kotlin/JS gradle plugin.

AlexanderPrendota:
Hello! Thanks for your PR. I need more time to have a look at it. 😊
(Currently, I'm working on creating a docker image and making configuration for starting the application as Amazon Lambda Function) 

Also, probably that feature might be interesting for @gzoritchak and @calvellido. (sorry for disturbing)

calvellido:
Indeed!  Thanks, let me know if I can lend a hand :+1:

bashor:
@rjaros could you please provide more about your usecases? 
Also, answer a few questions
1. How are you going to use it?
2. Should it be configurable by external users? 
3. If so, how and which parts?

bashor:
@gzoritchak @calvellido please feel free to share your thought too.

rjaros:
I would like to create an online playground app (kind of jsfiddle) for my framework (https://github.com/rjaros/kvision). I want to send a portion of Kotlin code from the user code editor to the service and get back a `bundle.js` file which can be used with a simple `index.html` file.

[...]

PR #37, opened 2020-01-27
Support for Webpack bundle (#33)

rjaros:
Support for bundling code with Webpack.

Requirements:
- working Node.js environment
- working Yarn installation

Gradle build creates additional directories in `META-INF` directory of JS executor. Requires working Yarn. If not available falls back to current behavior with a warning message.

New input project property `bundleType` with values:
- `none` (default, no bundling, current behavior)
- `plain` (Webpack processing without optimizations, fast but generates large output)
- `minimized` (Webpack processing with optimizations, generates minimized output but is quite slow)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -19,7 +19,7 @@ val copyDependencies by tasks.creating(Copy::class) {
     into(libJVMFolder)
 }
 val copyJSDependencies by tasks.creating(Copy::class) {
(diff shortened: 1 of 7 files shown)


dkrasnoff:
Thanks for contributing!
But this PR is closed, because there are no plans to support Webpack on Kotlin playground

PR #105, opened 2020-03-30
Streaming output

lyubortk:
Пока что без тестов


AlexanderPrendota:
Please rebase on master and resolve all conflicts

executors/src/main/kotlin/streaming/StreamingOutputMapper.kt
```diff
@@ -0,0 +1,47 @@
...
+  fun writeStdoutAsBytes(stdout: ByteArray): ByteArray = writeOutputAsBytes(stdout, "outStream")
+
+  fun writeStdoutAsBytes(stdout: String): ByteArray = writeStdoutAsBytes(stdout.toByteArray())
+
+  fun writeStderrAsBytes(stderr: ByteArray): ByteArray = writeOutputAsBytes(stderr, "errStream")
+
+  fun writeStderrAsBytes(stderr: String): ByteArray = writeStderrAsBytes(stderr.toByteArray())
+
+  fun writeTestRunInfoAsBytes(testRunInfo: TestRunInfo): ByteArray =
+    writeCustomObjectAsBytes(testRunInfo, "testResult")
```
AlexanderPrendota: Let's think later how we can avoid writing object fields as a raw string

src/main/kotlin/com/compiler/server/controllers/KotlinPlaygroundRestController.kt
```diff
@@ -56,4 +58,47 @@ class KotlinPlaygroundRestController(private val kotlinProjectExecutor: KotlinPr
...
+
+  /**
+   * An endpoint for kotlin playground requests with streaming output.
+   * Kotlin Playground component: https://github.com/JetBrains/kotlin-playground
+   */
+  @RequestMapping(
+    value = ["/kotlinServer"],
+    method = [RequestMethod.POST],
+    consumes = [MediaType.ALL_VALUE],
+    headers = ["Enable-Streaming=true"]
```
AlexanderPrendota: All your controllers get `headers = ["Enable-Streaming=true"]` should we consolidate ones

src/main/kotlin/com/compiler/server/executor/JavaStreamingExecutor.kt
```diff
@@ -0,0 +1,88 @@
...
+import java.util.concurrent.TimeUnit
+
+
+@Component
+class JavaStreamingExecutor(
+  private val streamingOutputMapper: StreamingOutputMapperComponent
+) {
+
+  companion object {
+    const val MAX_OUTPUT_SIZE = 100 * 1024
```
AlexanderPrendota: Let's store all restrictions for executer in one file

src/main/kotlin/com/compiler/server/executor/JavaStreamingExecutor.kt
```diff
@@ -0,0 +1,88 @@
...
+  fun execute(args: List<String>, output: OutputStream) {
+    Runtime.getRuntime().exec(args.toTypedArray()).apply {
+      try {
+        outputStream.close()
+        val standardOutput = streamingPipe(inputStream, output, MAX_OUTPUT_SIZE)
+
+        val currTime = System.currentTimeMillis()
+        val futuresList = Executors.newSingleThreadExecutor()
+          .invokeAll(listOf(standardOutput), EXECUTION_TIMEOUT, TimeUnit.MILLISECONDS)
+        println("Script execution time in millis: " + (System.currentTimeMillis() - currTime))
```
AlexanderPrendota: redundant =)

src/main/kotlin/com/compiler/server/executor/JavaStreamingExecutor.kt
```diff
@@ -0,0 +1,88 @@
...
+
+        val outputFuture = futuresList[0]
```
AlexanderPrendota:
```suggestion
        val outputFuture = futuresList.first()
```

src/main/kotlin/com/compiler/server/executor/JavaStreamingExecutor.kt
```diff
@@ -0,0 +1,88 @@
...
+        }
+      }
+    }
+  }
+
+  private fun streamingPipe(input: InputStream, output: OutputStream, limit: Int): Callable<PipeResult> = Callable {
+    var limitExceeded = false
+    try {
+      var bytesCopied = 0
+      val buffer = ByteArray(1024)
```
AlexanderPrendota: 1024 looks like a magic number. Let's move one to constant

src/main/kotlin/com/compiler/server/streaming/StreamingOutputMapperComponent.kt
```diff
@@ -0,0 +1,11 @@
+package com.compiler.server.streaming
+
+import com.compiler.server.model.ErrorDescriptor
+import executors.streaming.StreamingOutputMapper
+import org.springframework.stereotype.Component
+
+@Component
```
AlexanderPrendota: Why it should be a spring component?

src/main/kotlin/com/compiler/server/executor/JavaStreamingExecutor.kt
```diff
@@ -0,0 +1,88 @@
...
+import java.io.IOException
+import java.io.InputStream
+import java.io.OutputStream
+import java.lang.Integer.min
+import java.util.concurrent.Callable
+import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
+
+
+@Component
```
AlexanderPrendota: Why it should be a component? We can store streaming as an additional function.

So, why I have an objection about adding new spring components to the project: I try to avoid redundant component scanning because we execute that server as AWS lambda and we need to save time for starting the application

src/main/kotlin/com/compiler/server/streaming/StreamingOutputMapperComponent.kt
AlexanderPrendota: for example, we can use the extension function here `StreamingOutputMapper. writeErrorsAsBytes `

executors/src/main/kotlin/streaming/StreamingOutputMapper.kt
```diff
@@ -0,0 +1,47 @@
...
+    generator.writeStartObject()
+    generator.writeFieldName(fieldName)
+    mapper.writeValue(generator, obj)
+    generator.writeEndObject()
+    generator.close()
+    return DELIMITER + outputStream.toByteArray()
+  }
+
+  companion object {
+    private val DELIMITER = "\n\n".toByteArray()
```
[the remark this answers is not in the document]
lyubortk: To allow sending formatted jsons with newlines:
```
{
  "field1": "",
  "field2": ""
}
```
It is not really crucial but I believe that it might be useful to support parsing of human-readable jsons with formatting on a front-end side.

executors/src/main/kotlin/streaming/StreamingOutputMapper.kt
```diff
@@ -0,0 +1,47 @@
...
+  fun writeStderrAsBytes(stderr: ByteArray): ByteArray = writeOutputAsBytes(stderr, "errStream")
+
+  fun writeStderrAsBytes(stderr: String): ByteArray = writeStderrAsBytes(stderr.toByteArray())
+
+  fun writeTestRunInfoAsBytes(testRunInfo: TestRunInfo): ByteArray =
+    writeCustomObjectAsBytes(testRunInfo, "testResult")
+
+  private fun writeOutputAsBytes(output: ByteArray, fieldName: String): ByteArray {
+    val outputStream = ByteArrayOutputStream()
+    val generator = mapper.factory.createGenerator(outputStream)
```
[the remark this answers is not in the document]
lyubortk: generator has to be closed so i think it will be better if I use `.use{}` here

executors/src/main/kotlin/streaming/StreamingOutputMapper.kt
lyubortk: I didn't really understand what is the problem here. Those `write...AsBytes` functions that i wrote return byte arrays, not strings.

Or did you mean that we should move all field names to one place?

src/main/kotlin/com/compiler/server/executor/JavaStreamingExecutor.kt
lyubortk: Are you sure we don't want to use separate configurations for different executors? It can be useful to change settings for one executor without affecting the other one.

src/main/kotlin/com/compiler/server/executor/JavaStreamingExecutor.kt
lyubortk: I didn't know it could really affect performance. This solution could allow us not to instantiate OutputMapper multiple times (passing it around as a method parameter is even worse). I'll make it a regular class if it saves starting time.

src/main/kotlin/com/compiler/server/streaming/StreamingOutputMapperComponent.kt
lyubortk: I don't really want to use extension function here because it would mean that I have to expose `writeCustomObjectAsBytes` (I'd like to keep it `protected`). I'll just instantiate mapper as a field then (check out the commit i'm about to push)

src/main/kotlin/com/compiler/server/controllers/KotlinPlaygroundRestController.kt
lyubortk: > All your controllers get `headers = ["Enable-Streaming=true"]` should we consolidate ones

Not really, the old one gets `headers = ["Enable-Streaming!=true"]` but the new controller uses equals sign: `headers = ["Enable-Streaming=true"]`

executors/src/main/kotlin/streaming/StreamingOutputMapper.kt
lyubortk: > Or did you mean that we should move all field names to one place?

Anyway, i moved field names to companion object

lyubortk:
Github is notorious for screwing up comments on rebasing so I decided to merge actual master to this branch instead of rebasing. I can still rebase it later if it's necessary

executors/src/main/kotlin/streaming/StreamingOutputMapper.kt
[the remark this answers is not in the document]
AlexanderPrendota: +1

executors/src/main/kotlin/streaming/StreamingOutputMapper.kt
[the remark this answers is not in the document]
AlexanderPrendota: good =)

src/main/kotlin/com/compiler/server/controllers/KotlinPlaygroundRestController.kt
AlexanderPrendota: oh got it!
but is it possible to consolidate to controllers?

src/main/kotlin/com/compiler/server/executor/JavaStreamingExecutor.kt
AlexanderPrendota: yes, sure. the max output size and the timeout should be always the same in all executors

src/main/kotlin/com/compiler/server/executor/JavaStreamingExecutor.kt
AlexanderPrendota: So, it could be the kotlin object but we can combine two executors together let's discuss by phone about that

src/main/kotlin/com/compiler/server/compiler/components/KotlinCompiler.kt
```diff
@@ -42,19 +46,37 @@ class KotlinCompiler(
...
-      javaExecutor.execute(argsFrom(mainClass, output, arguments))
+      JavaExecutor.execute(argsFrom(mainClass, output, arguments))
         .asExecutionResult()
     }
   }
 
+  fun runStreaming(files: List<KtFile>, args: String, output: OutputStream) {
+    executeStreaming(files, output) { outputFilesDir, compiled, outputStream ->
+      val mainClass = JavaStreamingRunnerExecutor::class.java.name
+      val arguments = listOfNotNull(compiled.mainClass) + args.split(" ")
```
AlexanderPrendota, line 57: Let's use ` ` (space) as a constant for both run functions

src/main/kotlin/com/compiler/server/controllers/KotlinPlaygroundRestController.kt
```diff
@@ -56,4 +55,47 @@ class KotlinPlaygroundRestController(private val kotlinProjectExecutor: KotlinPr
     }
     return ResponseEntity.ok(result)
   }
+
+  /**
+   * An endpoint for kotlin playground requests with streaming output.
+   * Kotlin Playground component: https://github.com/JetBrains/kotlin-playground
+   */
+  @RequestMapping(
```
AlexanderPrendota, line 63: Missing API for the main controller

AlexanderPrendota:
Missing documentation for the new streaming API

AlexanderPrendota:
Use https://github.com/AlexanderPrendota/kotlin-compiler-server/pull/105 instead

PR #109, opened 2020-04-12
Support streaming output

lyubortk:


build.gradle.kts
```diff
@@ -88,12 +88,13 @@ fun buildPropertyFile() {
...
 }
 
-fun generateProperties(prefix: String = "") = """
+fun generateProperties(prefix: String = "", enableStreamingController: Boolean = true) = """
     # this file is autogenerated by build.gradle.kts
     kotlin.version=${kotlinVersion}
     policy.file=${prefix + policy}
     libraries.folder.jvm=${prefix + libJVMFolder}
     libraries.folder.js=${prefix + libJSFolder}
+    enable.streaming.controller=${enableStreamingController}
```
AlexanderPrendota, line 97: just `enable.streaming`. because we can re-use property somewhere else

src/main/kotlin/com/compiler/server/controllers/KotlinPlaygroundStreamingRestController.kt
```diff
@@ -0,0 +1,56 @@
...
+   */
+  @RequestMapping(
+    value = ["/kotlinServer"],
+    method = [RequestMethod.POST],
+    consumes = [MediaType.ALL_VALUE],
+    headers = ["Enable-Streaming=true"]
+  )
+  fun tryKotlinLangObsoleteEndpointStreaming(
+    @RequestParam type: String,
+    @RequestParam(required = false) project: Project?
```
AlexanderPrendota, line 28: Hm, we can make the project as not-null

dkrasnoff:
This PR is closed, because of outdated and will be continued in https://youtrack.jetbrains.com/issue/KTL-1021.

PR #110, opened 2020-04-13
Turn off transitive in copy dependencies

AlexanderPrendota:


build.gradle.kts
```diff
@@ -43,15 +45,18 @@ allprojects {
...
+    kotlinDependency("com.fasterxml.jackson.core:jackson-annotations:2.10.0")
+    // Kotlin libraries
     kotlinDependency("org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlinVersion")
+    kotlinDependency("org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlinVersion")
     kotlinDependency("org.jetbrains.kotlin:kotlin-reflect:$kotlinVersion")
-    kotlinDependency("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.5") {
-        exclude(group = "org.jetbrains.kotlin", module = "kotlin-stdlib")
-    }
+    kotlinDependency("org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion")
+    kotlinDependency("org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion")
```
dsavvinov, line 58: Am I missing something, or stdlilb is added twice?

build.gradle.kts
```diff
@@ -8,7 +8,9 @@ group = "com.compiler.server"
 version = "$kotlinVersion-SNAPSHOT"
 java.sourceCompatibility = JavaVersion.VERSION_1_8
 
-val kotlinDependency: Configuration by configurations.creating
+val kotlinDependency: Configuration by configurations.creating {
+    isTransitive = false
+}
 val kotlinJsDependency: Configuration by configurations.creating
```
dsavvinov, line 14: Is it intentional that we don't disable transitive dependencies for JS?

AlexanderPrendota: I have no idea about JS. @dsavvinov I guess kotlin-plugin does not affect JS dependencies. Shall I add?

build.gradle.kts
AlexanderPrendota: Oh, You are right. Will fix =)

build.gradle.kts
dsavvinov: I think we'd better disable them for JS as well, just for the sake of consistency (as far as I can see, it doesn't change anything currently; at most you'd need to add dependency to `kotlin-stdlib-common`)

AlexanderPrendota: Will add. thanks)

PR #120, opened 2020-04-24
Cors

AlexanderPrendota:


src/main/kotlin/com/compiler/server/configuration/CorsConfiguration.kt
```diff
@@ -0,0 +1,25 @@
+package com.compiler.server.configuration
+
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.web.cors.CorsConfiguration
+import org.springframework.web.cors.UrlBasedCorsConfigurationSource
+import org.springframework.web.filter.CorsFilter
+
+val ACCESS_CONTROL_ALLOW_ORIGIN_VALUE: String = System.getProperty("ACCESS_CONTROL_ALLOW_ORIGIN_VALUE", "*")
```
rvadim, line 9: Is `System.getProperty(...)` search environment variables? Or it's only jvm options?

AlexanderPrendota: ooops

PR #145, opened 2020-05-31
Kotless-based deployment for Play KotlinLang

TanVD:
The deployed version is available at `play.kotless.io`. Please, change parameters of deployment (in Gradle `kotless` closure) once merged.


AlexanderPrendota:
Did it in a local branch, will push

PR #158, opened 2020-07-22
Update KotlinFile.kt

zhelenskiy:
Bug fixed:
* Precondition: `null`-argument is provided to `expressionFor`.
* Expected behaviour: null returned
* Actual behaviour: infinite recursion

### DIFF
--- a/src/main/kotlin/com/compiler/server/compiler/KotlinFile.kt
+++ b/src/main/kotlin/com/compiler/server/compiler/KotlinFile.kt
@@ -34,7 +34,7 @@ class KotlinFile(val kotlinFile: KtFile) {
     (kotlinFile.viewProvider.document?.getLineStartOffset(line) ?: 0) + character
 
   private tailrec fun expressionFor(element: PsiElement?): PsiElement? =
-    if (element is KtExpression) element else expressionFor(element?.parent)
+    if (element is KtExpression?) element else expressionFor(element?.parent)
 
   companion object {
     fun from(project: Project, name: String, content: String) =


AlexanderPrendota:
Hey, big thanks for the PR. 😊 Could you please provide [a plain test](https://github.com/AlexanderPrendota/kotlin-compiler-server/blob/master/src/test/kotlin/com/compiler/server/JvmRunnerTest.kt) for that case?

zhelenskiy:
> Hey, big thanks for the PR.  Could you please provide [a plain test](https://github.com/AlexanderPrendota/kotlin-compiler-server/blob/master/src/test/kotlin/com/compiler/server/JvmRunnerTest.kt) for that case?

Oh, yes. But I found the bug a month ago when I had no response from the server for some input. I debugged and found infinite recursion here. Fixing it helped me.
However, I have just analyzed that such call couldn't have been done now (no such usages).
But there was a commit when you added the feature ([commit](https://github.com/AlexanderPrendota/kotlin-compiler-server/commit/7fc9b692bdc3e2a81c9f900ab16b629f9bbb2891#diff-d82c782c0de5f9167411da2e2b79058c)). So you have the feature that works incorrectly in (now) unused case. So:
* I can add the call for the test manually
* I can remove the ability to take nulls *(`T? -> T`)*
What would you like more?

AlexanderPrendota:
Hm...
> I can remove the ability to take nulls (T? -> T)

AFAIK if you return the PsiElement to a non-null value like `(element: PsiElement): PsiElement` some programs will produce an error because an element might be null. Unfortunately, I've forgotten to write a test for some reason and don't remember the output to reproduce =(

zhelenskiy:
> Hm...
> 
> > I can remove the ability to take nulls (T? -> T)
> 
> AFAIK if you return the PsiElement to a non-null value like `(element: PsiElement): PsiElement` some programs will produce an error because an element might be null. Unfortunately, I've forgotten to write a test for some reason and don't remember the output to reproduce =(

At least this cannot happen now because the only usage is `....?.invoke{ expressionFor(it) }`.
I think you changed smth somewhere and there is no need for the feature.
But infinite recursion is definitely an unpleasant result. So the current implementation of `null`-support couldn't solve the problem you had mentioned.

zhelenskiy:
> Hm...
> 
> > I can remove the ability to take nulls (T? -> T)
> 
> AFAIK if you return the PsiElement to a non-null value like `(element: PsiElement): PsiElement` some programs will produce an error because an element might be null. Unfortunately, I've forgotten to write a test for some reason and don't remember the output to reproduce =(

If you want to be able to handle null objects anywhere, I'll add a test with plain call of this method. Would you like it?

AlexanderPrendota:
Ah, got it. `invoke { expressionFor(it) }.` you are totally right. We are no need the null value over here.

AlexanderPrendota:
Let's use
> I can remove the ability to take nulls (T? -> T)

> But infinite recursion is definitely an unpleasant result. `tailrec` function does

zhelenskiy:
I accidentally created another [pull request](https://github.com/AlexanderPrendota/kotlin-compiler-server/pull/159).

AlexanderPrendota:
Hey, should we close this one?

zhelenskiy:
Yes, ofk

AlexanderPrendota:
@zhelenskiy big thanks for your help!

zhelenskiy:
> @zhelenskiy big thanks for your help!

You are welcome!!!

PR #173, opened 2020-08-03
feat(deps): add kotlin.test as dependency

zoobestik:


src/test/kotlin/com/compiler/server/DependenciesTest.kt
```diff
@@ -0,0 +1,22 @@
...
+@SpringBootTest
+class DependenciesTest {
+
+    @Autowired
+    private lateinit var testRunner: TestProjectRunner
+
+    @Test
+    fun `validate kotlin-test available`() {
+        testRunner.runWithException(
+                code = "import kotlin.test.assertTrue\n\nfun main(args: Array<String>) { assertTrue(false) }",
```
AlexanderPrendota, line 18: Please move to the [ExceptionProgramTest](https://github.com/AlexanderPrendota/kotlin-compiler-server/blob/master/src/test/kotlin/com/compiler/server/ExceptionInProgramTest.kt)

zoobestik: done

PR #231, opened 2020-12-06
Rewrite jvm indexation with declaration descriptors

PolinB:


indexation/src/main/kotlin/DescriptorsUtils.kt
```diff
@@ -55,6 +57,31 @@ internal fun DeclarationDescriptor.toImportInfo(): ImportInfo? {
   }
 }
 
+internal fun DeclarationDescriptor.getSubclassesAndAllStaticFunctions(): List<DeclarationDescriptor>? {
+  return when (this) {
```
AlexanderPrendota, line 61: why do u need "when" here? only one condition

indexation/src/main/kotlin/DescriptorsUtils.kt
```diff
@@ -55,6 +57,31 @@ internal fun DeclarationDescriptor.toImportInfo(): ImportInfo? {
...
+      } else null
+    }
+    else -> null
+  }
+}
+
+internal fun ModuleDescriptor.allImportsInfo(): List<ImportInfo> {
+  val packages = allPackages()
+  val imports = mutableListOf<ImportInfo>()
+  packages.forEach { fqName ->
```
AlexanderPrendota, line 75: val imports = packages.map{ ... }

indexation/src/main/kotlin/JsIndexationBuilder.kt
```diff
@@ -26,13 +24,7 @@ class JsIndexationBuilder(private val kotlinEnvironment: KotlinEnvironment): Ind
...
 
       configuration.moduleDescriptors.forEach { moduleDescriptor ->
-        val packages = moduleDescriptor.allPackages()
-        packages.forEach { fqName ->
-          val packageViewDescriptor = moduleDescriptor.getPackage(fqName)
-          val descriptors = packageViewDescriptor.memberScope
-            .getContributedDescriptors(DescriptorKindFilter.ALL, MemberScope.ALL_NAME_FILTER)
-          imports.addAll(descriptors.mapNotNull { it.toImportInfo() })
-        }
+        imports.addAll(moduleDescriptor.allImportsInfo())
```
AlexanderPrendota, line 27: vak imports = something.map { ... }, is it possible ?

indexation/src/main/kotlin/JvmIndexationBuilder.kt
```diff
@@ -1,163 +1,40 @@
...
-          file.listFiles { _: File?, name: String -> name.endsWith(JAR_EXTENSION) } ?: continue
-        for (jar in libFolderFiles) {
-          classPath.add(jar)
-        }
-      } else {
-        classPath.add(file)
-      }
-    }
-    return classPath.mapNotNull { it.toURI().toURL() }
+    File(outputFilename).writeText(jacksonObjectMapper().writerWithDefaultPrettyPrinter().writeValueAsString(getAllVariants()))
```
AlexanderPrendota, line 15: why do u need a pretty printer here?

indexation/src/main/kotlin/JvmIndexationBuilder.kt
```diff
@@ -1,163 +1,40 @@
...
-        val fullName = name.replace(File.separator, ".")
-        if (fullName.split(".").last() == MODULE_INFO_NAME) return@flatMap emptyList<ImportInfo>()
-        val clazz = runCatching { classLoader.loadClass(fullName) }.getOrNull()
-          ?: return@flatMap emptyList<ImportInfo>()
-        val classes = if (clazz.isKotlinClass()) {
-          allClassesFromKotlinClass(clazz)
-        } else {
-          allClassesFromJavaClass(clazz)
+  private fun getAllVariants(): List<ImportInfo> {
+    val imports = mutableListOf<ImportInfo>()
```
AlexanderPrendota, line 19: when I see the construction `val some = mutableListOf<ImportInfo>()` I always think we can replace one => something.map {}

AlexanderPrendota (changes requested):

indexation/src/main/kotlin/JsIndexationBuilder.kt
```diff
@@ -25,16 +15,8 @@ class JsIndexationBuilder(private val kotlinEnvironment: KotlinEnvironment): Ind
...
-        packages.forEach { fqName ->
-          val packageViewDescriptor = moduleDescriptor.getPackage(fqName)
-          val descriptors = packageViewDescriptor.memberScope
-            .getContributedDescriptors(DescriptorKindFilter.ALL, MemberScope.ALL_NAME_FILTER)
-          imports.addAll(descriptors.mapNotNull { it.toImportInfo() })
-        }
-      }
+      return@environment configuration.moduleDescriptors.flatMap { moduleDescriptor ->
+        moduleDescriptor.allImportsInfo()
+      }.distinct()
```
AlexanderPrendota, line 20: Is it possible to avoid a distinct call?

AlexanderPrendota (changes requested):

indexation/src/main/kotlin/JsIndexationBuilder.kt
PolinB: I believe that distinct call is needed, because without it, there are 100,000 more lines in the json file.

PR #248, opened 2021-02-07
fix: windows host process execution

wayfarer-rus:
When running under windows environment few characters messed up runtime

### DIFF
--- a/src/main/kotlin/com/compiler/server/compiler/components/KotlinCompiler.kt
+++ b/src/main/kotlin/com/compiler/server/compiler/components/KotlinCompiler.kt
@@ -128,9 +128,8 @@ class KotlinCompiler(
     outputDirectory: OutputDirectory,
     args: List<String>
   ): List<String> {
-    val pathSeparator = System.getProperty("path.separator") ?: ":"
     val classPaths = (kotlinEnvironment.classpath.map { it.absolutePath } + outputDirectory.path.toAbsolutePath().toString())
-      .joinToString(pathSeparator)
+      .joinToString(PATH_SEPARATOR)
     val policy = outputDirectory.path.resolve(policyFile.name).toAbsolutePath()
     return CommandLineArgument(
       classPaths = classPaths,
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/compiler/server/compiler/components/KotlinCompiler.kt
```diff
@@ -128,8 +128,9 @@ class KotlinCompiler(
     outputDirectory: OutputDirectory,
     args: List<String>
   ): List<String> {
+    val pathSeparator = System.getProperty("path.separator") ?: ":"
```
AlexanderPrendota, line 131: Looks like a global constant

AlexanderPrendota (changes requested):

src/main/kotlin/com/compiler/server/compiler/components/KotlinCompiler.kt
wayfarer-rus: where should I put it? 

your suggestion

AlexanderPrendota: private class field?

wayfarer-rus: done!

src/main/kotlin/com/compiler/server/compiler/components/KotlinCompiler.kt
```diff
@@ -148,4 +148,8 @@ class KotlinCompiler(
     }
   }
 
+  companion object {
+    private val PATH_SEPARATOR = System.getProperty("path.separator") ?: ":"
```
AlexanderPrendota, line 152: please top of the class

AlexanderPrendota (changes requested):

AlexanderPrendota:
@wayfarer-rus also maybe some test? =)

PR #267, opened 2021-03-29
Patch for 1.5.0-M2

anastasiiaSpaseeva:


src/main/kotlin/com/compiler/server/compiler/KotlinResolutionFacade.kt
```diff
@@ -17,9 +18,9 @@ import org.jetbrains.kotlin.resolve.lazy.BodyResolveMode
 
 @OptIn(FrontendInternals::class)
 class KotlinResolutionFacade(
-  override val project: Project,
-  override val moduleDescriptor: ModuleDescriptor,
-  private val componentProvider: ComponentProvider
+        override val project: Project,
+        override val moduleDescriptor: ModuleDescriptor,
```
AlexanderPrendota, line 22: return formatting

AlexanderPrendota (changes requested):

AlexanderPrendota:
the master must be only with a latest stable version

PR #269, opened 2021-03-31
Add JS IR compiler support

ilgonmic:
Initial support of JS IR compiler.
Possibly it can be improved in future (especially multi-file compilation instead one big bundle).

### DIFF
--- a/src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
+++ b/src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
@@ -4,24 +4,16 @@ import com.compiler.server.model.ErrorDescriptor
 import com.compiler.server.model.TranslationJSResult
 import com.compiler.server.model.toExceptionDescriptor
 import component.KotlinEnvironment
-import org.jetbrains.kotlin.cli.common.createPhaseConfig
 import org.jetbrains.kotlin.cli.common.messages.AnalyzerWithCompilerReport
-import org.jetbrains.kotlin.cli.common.messages.CompilerMessageSeverity
-import org.jetbrains.kotlin.cli.common.messages.CompilerMessageSourceLocation
-import org.jetbrains.kotlin.cli.common.messages.MessageCollector
-import org.jetbrains.kotlin.cli.js.K2JsIrCompiler
 import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
 import org.jetbrains.kotlin.ir.backend.js.MainModule
 import org.jetbrains.kotlin.ir.backend.js.compile
-import org.jetbrains.kotlin.ir.backend.js.jsPhases
-import org.jetbrains.kotlin.ir.backend.js.jsResolveLibraries
 import org.jetbrains.kotlin.js.config.JsConfig
 import org.jetbrains.kotlin.js.facade.K2JSTranslator
 import org.jetbrains.kotlin.js.facade.MainCallParameters
 import org.jetbrains.kotlin.js.facade.TranslationResult
 import org.jetbrains.kotlin.js.facade.exceptions.TranslationException
 import org.jetbrains.kotlin.psi.KtFile
-import org.jetbrains.kotlin.util.Logger
 import org.springframework.stereotype.Service
 import java.util.*
 
@@ -35,22 +27,19 @@ class KotlinToJSTranslator(
     private const val JS_CODE_BUFFER = "\nkotlin.kotlin.io.output.buffer;\n"
 
     private const val JS_IR_CODE_BUFFER = "moduleId.output._buffer;\n"
+
+    const val BEFORE_MAIN_CALL_LINE = 4
   }
 
   fun translate(
     files: List<KtFile>,
     arguments: List<String>,
-    coreEnvironment: KotlinCoreEnvironment,
-    useIrCompiler: Boolean
+    coreEnvironment: KotlinCoreEnvironment
   ): TranslationJSResult {
     val (errors, _) = errorAnalyzer.errorsFrom(files, coreEnvironment, isJs = true)
     return try {
       if (errorAnalyzer.isOnlyWarnings(errors)) {
-        (if (useIrCompiler) {
-          doTranslateWithIr(files, arguments, coreEnvironment)
-        } else {
-          doTranslate(files, arguments, coreEnvironment)
-        }).also {
+        doTranslate(files, arguments, coreEnvironment).also {
           it.addWarnings(errors)
         }
       } else {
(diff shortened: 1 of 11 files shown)


src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
```diff
@@ -22,13 +33,24 @@ class KotlinToJSTranslator(
...
+
+    private const val JS_IR_CODE_BUFFER = "moduleId.output._buffer;\n"
   }
 
-  fun translate(files: List<KtFile>, arguments: List<String>, coreEnvironment: KotlinCoreEnvironment): TranslationJSResult {
+  fun translate(
+    files: List<KtFile>,
+    arguments: List<String>,
+    coreEnvironment: KotlinCoreEnvironment,
+    useIrCompiler: Boolean
```
AlexanderPrendota, line 44: I would suggest using another method called `translateIR()` instead of boolean flag

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
```diff
@@ -73,4 +95,66 @@ class KotlinToJSTranslator(
...
+      propertyLazyInitialization = false,
+      mainArguments = arguments
+    )
+    val jsCode = result.jsCode!!.mainModule
+
+    val listLines = jsCode
+      .lineSequence()
+      .toMutableList()
+
+    listLines.add(listLines.size - 4, "if (isRewrite) output = new BufferedOutput_0()")
```
AlexanderPrendota, line 154: listLines.size - 4
listLines.size - 1

looks like the magic numbers. Please add the comment on why we need such actions. For my point of view it looks like a specific hack

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
```diff
@@ -73,4 +95,66 @@ class KotlinToJSTranslator(
...
+        location: CompilerMessageSourceLocation?
+      ) {
+      }
+    }
+    val result = compile(
+      currentProject,
+      MainModule.SourceFiles(files),
+      AnalyzerWithCompilerReport(configuration.configuration),
+      configuration.configuration,
+      createPhaseConfig(jsPhases, K2JsIrCompiler().createArguments(), messageCollector),
```
AlexanderPrendota, line 142: Do we need to create K2JsIrCompiler().createArguments() each time when we have a user request?

AlexanderPrendota: the same questions for other configuration parameters

src/main/kotlin/com/compiler/server/model/Project.kt
```diff
@@ -17,5 +17,11 @@ enum class ProjectType(@JsonValue val id: String) {
   JAVA("java"),
   JUNIT("junit"),
   CANVAS("canvas"),
-  JS("js")
+  JS("js"),
+  JS_IR("js-ir");
+
+  fun isJsRelated(): Boolean =
```
AlexanderPrendota, line 23: inline 

`=  this == JS || this == JS_IR ||  this == CANVAS`

src/test/kotlin/com/compiler/server/CommandLineArgumentsTest.kt
```diff
@@ -25,19 +25,39 @@ class CommandLineArgumentsTest : BaseExecutorTest() {
 
   @Test
   fun `command line arguments js test`() {
+    commandLineArgumentsJsCommon(useIrCompiler = false)
+  }
+
+  @Test
+  fun `command line arguments js ir test`() {
+    commandLineArgumentsJsCommon(useIrCompiler = true)
```
AlexanderPrendota, line 33: if we want to have the same tests with plain JS compiler we don't need to write additional functions. Let's send the 2 requests per each JS test: 

1) for plain compiler
2) for ir 

and check results one by one

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
```diff
@@ -73,4 +95,66 @@ class KotlinToJSTranslator(
       TranslationJSResult(errors = errors)
     }
   }
+
+  @Throws(TranslationException::class)
+  private fun doTranslateWithIr(
```
AlexanderPrendota, line 100: Do we have differences between IR and PLAIN compilers? We have an "indexes" module to build JSON indexes for auto-imports and competitions

src/test/kotlin/com/compiler/server/CommandLineArgumentsTest.kt
AlexanderPrendota: also I would suggest using function instead of boolean parameter =)

AlexanderPrendota (changes requested):

src/test/kotlin/com/compiler/server/CommandLineArgumentsTest.kt
ilgonmic: There is test (with `canvas`) which actual only for old compiler for now.
So we can use boolean parameter to distinguish, but it requires copy-paste in "ir" and "old" methods

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
ilgonmic: It is not actual for stdlib, because stdlib contains both `ir` and `old` compiler, so it can be ok to use `stdlib` to get indices from "old" compiler and use it for IR.

src/main/kotlin/com/compiler/server/model/Project.kt
ilgonmic: Ready

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
ilgonmic: Ready

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
ilgonmic: Ready

src/test/kotlin/com/compiler/server/CommandLineArgumentsTest.kt
ilgonmic: Ready

AlexanderPrendota (approved):

PR #279, opened 2021-04-14
add 1.5.0-RC

woainikk:


src/test/kotlin/com/compiler/server/KotlinFeatureSince150.kt
```diff
@@ -6,62 +6,146 @@ import org.junit.jupiter.api.Test
 
 class KotlinFeatureSince150 : BaseExecutorTest() {
     @Test
-    fun `KT-45991 on JVM IR BE`() {
+    fun `unsigned integer types`() {
         run(
-                code = """
-                fun ex1_(res: Result<Int>) {
```
AlexanderPrendota, line 12: why did u delete the test?

versions.json
```diff
@@ -14,5 +14,8 @@
   },
   {
     "version": "1.5.0-M2"
```
AlexanderPrendota, line 16: do u need 1.5.0 -M2?

PR #342, opened 2021-08-19
KTL-89 JSON events in the stdout

nikpachoo:
Hi, @AlexanderPrendota!
YouTrack issue: [KTL-89](https://youtrack.jetbrains.com/issue/KTL-89)

We want to observe compiler-server errors with the ELK stack.
For that, the log's event should be in JSON format and contain some fields of request and response.

So, there is a couple of changes to achieve it:
- Format logs event as JSON (3f0d0ff)
- Log execution errors by KotlinProjectExecutor (360bc98)

For example, now the event with compilation errors look like:
```
{
  "@timestamp": "2021-08-26T18:38:57.875+03:00",
  "@version": "1",
  "message": "Compilation end with errors",
  "logger_name": "com.compiler.server.service.KotlinProjectExecutor",
  "thread_name": "http-nio-8080-exec-8",
  "level": "INFO",
  "level_value": 20000,
  "errors": [
    "No main method found in project.\n"
  ],
  "confType": "JAVA",
  "kotlinVersion": "1.5.21"
}
```

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -8,6 +8,7 @@ val kotlinIdeVersion: String by System.getProperties()
 val policy: String by System.getProperties()
 val indexes: String by System.getProperties()
 val indexesJs: String by System.getProperties()
+val executorLogs: String by System.getProperties()
 
 group = "com.compiler.server"
 version = "$kotlinVersion-SNAPSHOT"
@@ -129,6 +130,7 @@ fun generateProperties(prefix: String = "") = """
     indexesJs.file=${prefix + indexesJs}
     libraries.folder.jvm=${prefix + libJVMFolder}
     libraries.folder.js=${prefix + libJSFolder}
+    executor.logs=${executorLogs}
 """.trimIndent()
 
 tasks.withType<KotlinCompile> {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -89,6 +89,7 @@ dependencies {
     implementation("org.springframework.boot:spring-boot-starter-web")
     implementation("com.amazonaws.serverless:aws-serverless-java-container-springboot2:1.6")
     implementation("junit:junit:4.12")
+    implementation("net.logstash.logback:logstash-logback-encoder:6.6")
     implementation("org.jetbrains.intellij.deps:trove4j:1.0.20200330")
     implementation("org.jetbrains.kotlin:kotlin-reflect:$kotlinVersion")
     implementation("org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion")
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,3 +3,4 @@ systemProp.kotlinIdeVersion=1.5.30-release-412
 systemProp.policy=executor.policy
 systemProp.indexes=indexes.json
 systemProp.indexesJs=indexesJs.json
+systemProp.executorLogs=true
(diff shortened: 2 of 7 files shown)


build.gradle.kts
```diff
@@ -88,6 +88,8 @@ dependencies {
     implementation("org.springframework.boot:spring-boot-starter-web")
     implementation("com.amazonaws.serverless:aws-serverless-java-container-springboot2:1.6")
     implementation("junit:junit:4.12")
+    implementation("ch.qos.logback:logback-access:1.2.5")
```
AlexanderPrendota, line 91: Why not use the default spring logger?

build.gradle.kts
```diff
@@ -125,6 +127,8 @@ fun generateProperties(prefix: String = "") = """
     indexesJs.file=${prefix + indexesJs}
     libraries.folder.jvm=${prefix + libJVMFolder}
     libraries.folder.js=${prefix + libJSFolder}
+    server.error.include-message=always
+    spring.main.banner-mode=off
```
AlexanderPrendota, line 131: Why? =)

AlexanderPrendota (changes requested):

src/main/kotlin/com/compiler/server/model/ExecutionResult.kt
```diff
@@ -14,6 +14,34 @@ open class ExecutionResult(
...
+        }
+      }
+    }
+
+    return messages
+  }
+
+  private fun textWithError() = text.startsWith("<errStream>")
+
+  private fun parseTextError(): String? = Regex(
```
AlexanderPrendota, line 41: we can put message with <errStream> :) why not?
also:
```
string.removePrefix("bla")
string.removeSuffix("bla")

File: src/main/kotlin/com/compiler/server/service/KotlinProjectExecutor.kt [hunk 4]
```diff
@@ -95,4 +96,24 @@ class KotlinProjectExecutor(
   private fun getFilesFrom(project: Project, coreEnvironment: KotlinCoreEnvironment) = project.files.map {
     KotlinFile.from(coreEnvironment.project, it.name, it.text)
   }
-}
\ No newline at end of file
+
+  private fun logExecutionErrors(project: Project, executionResult: ExecutionResult) {
```
AlexanderPrendota, line 100: The method should be outsize of this class. Probably in utils package

AlexanderPrendota (changes requested):

src/main/kotlin/com/compiler/server/model/ExecutionResult.kt
```diff
@@ -14,6 +14,19 @@ open class ExecutionResult(
...
   }
+
+  fun isUnsuccessful() =
+    textWithError() || errors.any { (_, value) -> value.any { it.severity == ProjectSeveriry.ERROR } }
+
+  fun errorsMessages(): List<String> {
+    val compilationErrors: List<String> =
+      errors.flatMap { it.value }.filter { it.severity == ProjectSeveriry.ERROR }.map { it.message }
+
+    return if (textWithError()) listOf(text) + compilationErrors
```
AlexanderPrendota, line 25: no need to wrap the text. Example: https://pl.kotl.in/Y_8DyPMz4

AlexanderPrendota (changes requested):

gradle.properties
```diff
@@ -3,3 +3,4 @@ systemProp.kotlinIdeVersion=1.5.30-release-412
 systemProp.policy=executor.policy
 systemProp.indexes=indexes.json
 systemProp.indexesJs=indexesJs.json
+systemProp.executorLogs=true
```
AlexanderPrendota, line 6: Documentation for this in Readme. Please please :)

AlexanderPrendota (changes requested):

PR #356, opened 2021-09-04
Introduce logger for output

AlexanderPrendota:
Main PR: https://github.com/AlexanderPrendota/kotlin-compiler-server/pull/342


AlexanderPrendota:
cc @nikpachoo

PR #357, opened 2021-09-06
chore: update to intellij 211 platform

woainikk:


gradle.properties
```diff
@@ -1,5 +1,5 @@
 systemProp.kotlinVersion=1.5.30
```
AlexanderPrendota, line 1: why kotlin version 1530 but ide version 1531?

AlexanderPrendota (changes requested):

gradle.properties
woainikk: kotlin 1.5.31 is not released yet

PR #361, opened 2021-09-13
chore(deps): update deps related to bintray

nikpachoo:
I replaced the bintray links with the current ones that I got from the [master branch](https://github.com/AlexanderPrendota/kotlin-compiler-server/blob/master/build.gradle.kts#L56).

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -52,9 +52,10 @@ allprojects {
     repositories {
         mavenCentral()
         maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/kotlin-ide")
-        maven("https://kotlin.bintray.com/kotlin-ide-plugin-dependencies")
-        maven("https://jetbrains.bintray.com/intellij-third-party-dependencies")
-        maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/kotlin-compiler-for-ide")
+        maven("https://cache-redirector.jetbrains.com/jetbrains.bintray.com/intellij-third-party-dependencies")
+        maven("https://maven.pkg.jetbrains.space/kotlin/p/kotlin/kotlin-ide-plugin-dependencies")
+        maven("https://www.myget.org/F/rd-snapshots/maven/")
+        maven("https://kotlin.jetbrains.space/p/kotlin/packages/maven/kotlin-ide")
     }
     afterEvaluate {
         dependencies {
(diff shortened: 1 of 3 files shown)


AlexanderPrendota:
@woainikk @anastasiiaSpaseeva Folks, please validate deps

build.gradle.kts
```diff
@@ -95,7 +96,7 @@ dependencies {
     implementation("org.jetbrains.kotlin:kotlin-compiler:$kotlinVersion")
     implementation("org.jetbrains.kotlin:kotlin-script-runtime:$kotlinVersion")
     implementation("org.jetbrains.kotlin:kotlin-stdlib-js:$kotlinVersion")
-    implementation("org.jetbrains.kotlin:kotlin-compiler-for-ide:$kotlinVersion")
+    implementation("org.jetbrains.kotlin:kotlin-compiler-for-ide:$kotlinVersion-release-310")
```
AlexanderPrendota, line 99: I think it;s a bad idea to pass raw string "-release-310" to deps. We always update dependencies via gradle.properties file

common/build.gradle.kts
```diff
@@ -7,5 +7,10 @@ plugins {
 
 dependencies {
     implementation("org.jetbrains.kotlin:kotlin-compiler:$kotlinVersion")
+    implementation("org.jetbrains.kotlin:kotlin-compiler-for-ide") {
+        version {
+            strictly( "$kotlinVersion-release-310")
```
AlexanderPrendota, line 12: why do u need such changes?

AlexanderPrendota (changes requested):

build.gradle.kts
woainikk: there should be used $kotlinIdeVersion

common/build.gradle.kts
nikpachoo: The problem is that the dependency `org.jetbrains.kotlin:common:202-$kotlinIdeVersion-IJ8194.7` refers to "org.jetbrains.kotlin:kotlin-compiler-for-ide-1.4.30" which cannot be resolved cause there is no such version ([mvnrepository](https://mvnrepository.com/artifact/org.jetbrains.kotlin/kotlin-compiler-for-ide)).

woainikk: why kotlin 1.4.30 version used, not the new one (1.5.30)?

build.gradle.kts
AlexanderPrendota: riiiiiight

AlexanderPrendota:
ah exactly, the PR related to kotlin 1.4.30

common/build.gradle.kts
AlexanderPrendota: because the pr into 1430)

build.gradle.kts
nikpachoo: I've tried to use $kotlinIdeVersion, but it didn't work. I got an error:
```
Could not find org.jetbrains.kotlin:kotlin-compiler-for-ide:1.4.30-release.
```

AlexanderPrendota: kotlinVersion=1.4.30-release-310 ?

AlexanderPrendota (changes requested):

build.gradle.kts
nikpachoo: > kotlinVersion=1.4.30-release-310
nope, because the other dependencies that use this variable stop working.

woainikk (approved):

PR #389, opened 2021-11-15
fix: change output directory path for Lambda

nikpachoo:
For Lambda's function temporary files should be written to the /tmp directory.


build.gradle.kts
```diff
@@ -130,6 +130,7 @@ fun generateProperties(prefix: String = "") = """
     libraries.folder.jvm=${prefix + libJVMFolder}
     libraries.folder.js=${prefix + libJSFolder}
     executor.logs=${executorLogs}
+    outputDirectoryPath=${outputDirectoryPath}
```
AlexanderPrendota, line 133: no need to write outputDirectoryPath, let's ignore such case

AlexanderPrendota (changes requested):

src/main/kotlin/com/compiler/server/compiler/components/KotlinCompiler.kt
```diff
@@ -36,7 +36,8 @@ class KotlinCompiler(
   private val kotlinEnvironment: KotlinEnvironment,
   private val javaExecutor: JavaExecutor,
   private val librariesFile: LibrariesFile,
-  @Value("\${policy.file}") private val policyFileName: String
+  @Value("\${policy.file}") private val policyFileName: String,
+  @Value("\${outputDirectoryPath:}") private val outputDirectoryPath: String
```
nikpachoo, line 40: As I understand it, there is no way to specify `@Value` as optional, so I had to add the default value with the colon.

AlexanderPrendota: property can be null

build.gradle.kts
```diff
@@ -130,6 +130,7 @@ fun generateProperties(prefix: String = "") = """
     libraries.folder.jvm=${prefix + libJVMFolder}
     libraries.folder.js=${prefix + libJSFolder}
     executor.logs=${executorLogs}
+    ${if (outputDirectoryPath.isNotEmpty()) "outputDirectoryPath=${outputDirectoryPath}" else ""}
```
AlexanderPrendota, line 133: extract a variable instead

src/main/kotlin/com/compiler/server/compiler/components/KotlinCompiler.kt
```diff
@@ -106,7 +107,12 @@ class KotlinCompiler(
     val dir = System.getProperty("user.dir")
     val libDir = librariesFile.jvm.absolutePath
     val sessionId = UUID.randomUUID().toString().replace("-", "")
-    val outputDir = Paths.get(dir, "tmp", sessionId)
+
+    val outputDir = if (outputDirectoryPath.isNotEmpty()) {
```
AlexanderPrendota, line 111: if (a != null)

AlexanderPrendota (changes requested):

src/main/kotlin/com/compiler/server/compiler/components/KotlinCompiler.kt
zoobestik: Why we don't just use `System.getProperty("java.io.tmpdir")` temporary resources? It will works for all environments without condition. No?

AlexanderPrendota: i like it

nikpachoo: I have not use it, because I didn't know about it. 
Is it work with Lambda properly?

zoobestik:
@nikpachoo, @AlexanderPrendota,

I'm still think that's part is not good, cause some of server exception shows like `ExecutionResult` it's not user frendly, it's potentially usecure show internal output for user. 
https://github.com/AlexanderPrendota/kotlin-compiler-server/blob/4a16f723a44ce0158514338892543b1aad453e2a/src/main/kotlin/com/compiler/server/compiler/components/KotlinCompiler.kt#L81-L102

AlexanderPrendota:
@zoobestik true. we can log with error server error and show to users some message

src/main/kotlin/com/compiler/server/compiler/components/KotlinCompiler.kt
nikpachoo: tested with Lambda, works fine

PR #477, opened 2022-05-13
Eagerly initialized output

ilgonmic:
Since 1.6.20, Kotlin/JS uses lazy initialization of top level properties.
So, to rewrite `output` top level property, it should be initialized firstly.

### DIFF
--- a/src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
+++ b/src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
@@ -31,6 +31,13 @@ class KotlinToJSTranslator(
 
     private const val JS_IR_CODE_BUFFER = "moduleId.output._buffer;\n"
 
+    private val JS_IR_OUTPUT_REWRITE = """
+        if (kotlin.isRewrite) {
+            init_properties_console_kt_3903220573();
+            output = new BufferedOutput_0()
+        }
+        """.trimIndent()
+
     const val BEFORE_MAIN_CALL_LINE = 4
   }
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
```diff
@@ -129,7 +129,15 @@ class KotlinToJSTranslator(
       .lineSequence()
       .toMutableList()
 
-    listLines.add(listLines.size - BEFORE_MAIN_CALL_LINE, "if (kotlin.isRewrite) output = new BufferedOutput_0()")
+    listLines.add(
+      listLines.size - BEFORE_MAIN_CALL_LINE,
+      """
+        if (kotlin.isRewrite) {
```
AlexanderPrendota, line 135: maybe it should be a constant?

AlexanderPrendota (changes requested):

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
ilgonmic: Fair, fixed

AlexanderPrendota (approved):

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
AlexanderPrendota: @ilgonmic do u need a test for it?

ilgonmic: It is a bit hard, because it requires js execution for useful test to be sure, that correct init function is used.

PR #501, opened 2022-06-16
feat: add an option to translate JS_IR with DCE.

JSMonk:
Add an query argument for adding dead code elimination step before IR to JS translating

### DIFF
--- a/src/main/kotlin/com/compiler/server/service/KotlinProjectExecutor.kt
+++ b/src/main/kotlin/com/compiler/server/service/KotlinProjectExecutor.kt
@@ -51,8 +51,10 @@ class KotlinProjectExecutor(
     return convertJsWithConverter(project, kotlinToJSTranslator::doTranslate)
   }
 
-  fun convertToJsIr(project: Project): TranslationJSResult {
-    return convertJsWithConverter(project, kotlinToJSTranslator::doTranslateWithIr)
+  fun convertToJsIr(project: Project, shouldEliminateDeadCode: Boolean): TranslationJSResult {
+    return convertJsWithConverter(project) { files, args, env ->
+       kotlinToJSTranslator.doTranslateWithIr(files, args, env, shouldEliminateDeadCode)
+    }
   }
 
   fun complete(project: Project, line: Int, character: Int): List<Completion> {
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/compiler/server/service/KotlinProjectExecutor.kt
```diff
@@ -51,8 +51,10 @@ class KotlinProjectExecutor(
     return convertJsWithConverter(project, kotlinToJSTranslator::doTranslate)
   }
 
-  fun convertToJsIr(project: Project): TranslationJSResult {
-    return convertJsWithConverter(project, kotlinToJSTranslator::doTranslateWithIr)
+  fun convertToJsIr(project: Project, shouldEliminateDeadCode: Boolean): TranslationJSResult {
```
nikpachoo, line 54: Please, fix the usages of the method or set a default value for "shouldEliminateDeadCode".
[BaseExecutorTest](https://github.com/AlexanderPrendota/kotlin-compiler-server/blob/492f081d8144af5e033694f565f555d2a6f3841d/src/test/kotlin/com/compiler/server/base/BaseExecutorTest.kt#L55-L62) will not pass due to the method's signature mismatch.

AlexanderPrendota: also it;s good to have additional test for that feature

src/test/kotlin/com/compiler/server/base/BaseExecutorTest.kt
```diff
@@ -50,16 +50,33 @@ class BaseExecutorTest {
...
     contains: String
   ) = testRunner.multiRunJs(code, contains) { project ->
-    convertToJsIr(project)
+    convertToJsIr(project, shouldEliminateDeadCode = false)
+  }
+
+  fun runJsIrWithDce(
+    code: String,
+    contains: String = "",
+    notContain: String = "",
```
AlexanderPrendota, line 68: notContain`s` mb?

AlexanderPrendota (changes requested):

AlexanderPrendota:
@JSMonk do u need it?

dkrasnoff:
UPD: Artem said this PR is still relevant and will be updated.

PR #508, opened 2022-07-04
fix(Kotlin/JS): replace deprecated Kotlin/JS IR compiler API with a n…

JSMonk:
The PR provides a new way to compile JS with IR.
The currently used API will be removed from the compiler soon, so the provided changes will prevent breaks on the compiler API side.

### DIFF
--- a/src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
+++ b/src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
@@ -9,8 +9,10 @@ import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
 import org.jetbrains.kotlin.ir.backend.js.CompilerResult
 import org.jetbrains.kotlin.ir.backend.js.compile
 import org.jetbrains.kotlin.ir.backend.js.prepareAnalyzedSourceModule
+import org.jetbrains.kotlin.ir.backend.js.transformers.irToJs.TranslationMode
 import org.jetbrains.kotlin.ir.backend.js.transformers.irToJs.IrModuleToJsTransformer
-import org.jetbrains.kotlin.ir.declarations.impl.IrFactoryImpl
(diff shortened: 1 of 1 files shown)


AlexanderPrendota:
should I merge it? =)

JSMonk:
Sure. We've just started to delete the described API, so, it will be great if we will merge it before compiler updates that will not contain deprecated API.

AlexanderPrendota:
Okay, do we need to have test for such changes?

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
```diff
@@ -115,19 +117,25 @@ class KotlinToJSTranslator(
...
+      dce = false,
+      perModule = false,
+      minimizedMemberNames = false
     )
 
-    val compiledModule: CompilerResult = transformer.generateModule(ir.allModules)
-    val jsCode = compiledModule.outputs.values.single().jsCode
+    val compiledModule: CompilerResult = transformer.generateModule(ir.allModules, setOf(translationMode))
+
+    val jsCode = compiledModule.outputs[translationMode]!!.jsCode
```
AlexanderPrendota, line 138: is it possible to avoid using `!!`?

AlexanderPrendota (changes requested):

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
AlexanderPrendota: cc @JSMonk

AlexanderPrendota:
@JSMonk ping

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
AlexanderPrendota: yes, let's do it

AlexanderPrendota:
@JSMonk friendly ping

PR #540, opened 2022-09-06
KTL-114 Add kotlinx.serialization compiler plugin to Kotlin Playground

nikpachoo:
There is a draft for supporting kotlinx.coroutines.

Status note:
These changes are not working for some reason. I am trying to find out what is wrong.

### DIFF
--- a/common/src/main/kotlin/component/KotlinEnvironment.kt
+++ b/common/src/main/kotlin/component/KotlinEnvironment.kt
@@ -30,7 +30,8 @@ import java.io.File
 
 class KotlinEnvironment(
   val classpath: List<File>,
-  additionalJsClasspath: List<File>
+  additionalJsClasspath: List<File>,
+  private val compilerPlugins: List<File>
 ) {
   companion object {
     /**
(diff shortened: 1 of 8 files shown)


common/src/main/kotlin/component/KotlinEnvironment.kt
```diff
@@ -113,6 +114,11 @@ class KotlinEnvironment(
   private fun createConfiguration(): CompilerConfiguration {
     val arguments = K2JVMCompilerArguments()
     parseCommandLineArguments(additionalCompilerArguments, arguments)
+
+    if (compilerPlugins.isNotEmpty()) {
```
AlexanderPrendota, line 118: is it possible to pass arguments only if we need plugins?

AlexanderPrendota (changes requested):

common/src/main/kotlin/component/KotlinEnvironment.kt
nikpachoo: Sorry, I am not sure what you mean. 
The arguments are passed only if the plugin was defined if the build.gradle.kts.
So, for me, it is what you asked for :)

AlexanderPrendota: it's not necessary to have compiler plugins if we don't use them in the code

PR #568, opened 2022-11-14
chore(release): add kotlin 1.8.0-Beta

woainikk:


src/test/resources/test-compile-data/jvm/kotlin-web-site/whatsnew15/f581dd74809f95d6448c98641a1ef376.2.kt
```diff
@@ -1,10 +1,11 @@
 import kotlin.time.Duration
+import kotlin.time.Duration.Companion.milliseconds
 import kotlin.time.ExperimentalTime
 
 @ExperimentalTime
 fun main() {
 //sampleStart
-    val duration = Duration.milliseconds(120000)
+    val duration = 120000.milliseconds
```
AlexanderPrendota, line 8: why do u need to fix the such file? it's autogenerated. Please fix the sample on kotlin-web-site

AlexanderPrendota (changes requested):

src/test/resources/test-compile-data/jvm/kotlin-web-site/whatsnew15/f581dd74809f95d6448c98641a1ef376.2.kt
woainikk: kotlin library api has changed in 1.8.0, the old method is deprecated and without the change the test will fail. In 1.5 (as mentioned in sample on kotlin-web-site) this code works

PR #575, opened 2022-11-30
Kotlin 1.8.0-RC

woainikk:


build.gradle.kts
```diff
@@ -46,15 +46,16 @@ val copyJSDependencies by tasks.creating(Copy::class) {
 plugins {
     id("org.springframework.boot") version "2.7.6"
     id("io.spring.dependency-management") version "1.1.0"
-    kotlin("jvm") version "1.8.0-Beta"
-    kotlin("plugin.spring") version "1.8.0-Beta"
+    kotlin("jvm") version "1.8.0-RC"
```
AlexanderPrendota, line 49: I think we should use plugins from default (latest-stable) version in master

woainikk: fixed

AlexanderPrendota: afaik stable is 1.7.21

PR #604, opened 2023-02-10
Ilgonmic/js ir ic

ilgonmic:
It adds support of compiler backend incremental compilation for Kotlin JS IR.
It makes compilation much faster than before.

It makes some prepared cache, stored it and then on request, it copied to temp directory and using during compilation process.

Additionally compressing on server is enabled to make sending scripts less

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,7 @@ FROM openjdk:11.0.16-jdk as build
 
 ENV KOTLIN_LIB=1.8.20-Beta
 ENV KOTLIN_LIB_JS=1.8.20-Beta-js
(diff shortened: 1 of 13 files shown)


ilgonmic:
Tests on JS IR already exists
https://github.com/AlexanderPrendota/kotlin-compiler-server/blob/master/src/test/kotlin/com/compiler/server/ConvertToJsRunnerTest.kt#L36
I changed implementation details, that it should use incremental cache.

Or you mean some specific tests that incrementality work? I can think about it

AlexanderPrendota (approved):

PR #611, opened 2023-02-24
Use kotlin-test-junit

hfhbd:
Kotlin test junit contains the kotlin.test.Test typealias to org.junit.Test.
This change allows you to write a MPP test without explicit junit Test annotation:
```kotlin
import kotlin.test.*

class Testing
  @Test fun testing() {
    assertTrue(true) 
  }
}
```

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -74,6 +74,7 @@ allprojects {
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


AlexanderPrendota:
Hi, thank you. I suppose we need a test for it :) Just some code snippet to make sure we don't lose such behavior.

PR #618, opened 2023-03-18
Prevent stray output from reaching stdout

phiSgr:
prevents prints that happen after (user's) `main` `return`s from polluting stdout.

Context: https://stackoverflow.com/questions/75714151/kotlin-coroutines-order-of-printing-logs/75760394#75760394


phiSgr:
On second thought, not switching System.out back may cause stray prints become data races as the main thread reads from `ByteArrayOutputStream` and the thread the prints writes to it.

But if that is the reason for calling `setOut`, `System.err` should also be switched back.

---

Or if the writes synchronize over the BAOS as I suggested in #619, the data race can be prevented by

```kt
synchronized(outputStream) {
    outputObj.text = outputStream.toString()
        .replace("</errStream><errStream>".toRegex(), "")
        .replace("</outStream><outStream>".toRegex(), "")
}
```

AlexanderPrendota:
Hi, is it possible to add a test snippet?

src/test/kotlin/com/compiler/server/CoroutinesRunnerTest.kt
```diff
@@ -57,10 +57,11 @@ class CoroutinesRunnerTest : BaseExecutorTest() {
 
   @Test
   fun `base coroutines test 7`() {
-    run(
+    val result = run(
       code = "import kotlinx.coroutines.*\n\nfun main() = runBlocking {\nGlobalScope.launch {\n    repeat(1000) { i ->\n        println(\"I'm sleeping \$i ...\")\n        delay(500L)\n    }\n}\ndelay(1300L) // just quit after delay    \n}",
-      contains = "I'm sleeping 0 ...\nI'm sleeping 1 ...\nI'm sleeping 2 ...\n"
+      contains = ""
     )
+    Assertions.assertEquals("<outStream>I'm sleeping 0 ...\nI'm sleeping 1 ...\nI'm sleeping 2 ...\n</outStream>", result.text)
```
phiSgr, line 64: If you go before the fix with `git checkout 5e25435`, this test will intermittently fail.

src/main/kotlin/com/compiler/server/model/ProgramOutput.kt
```diff
@@ -26,14 +26,7 @@ data class ProgramOutput(
       standardOutput.isBlank() -> ExecutionResult()
       else -> {
         try {
-          // coroutines can produced incorrect output. see example in `base coroutines test 7`
-          if (standardOutput.startsWith("{")) outputMapper.readValue(standardOutput, ExecutionResult::class.java)
-          else {
-            val result = outputMapper.readValue("{" + standardOutput.substringAfter("{"), ExecutionResult::class.java)
```
phiSgr, line 32: I believe we won't need this workaround, but I wouldn't object to keeping it.

executors/src/main/kotlin/JUnitExecutors.kt
```diff
@@ -29,13 +29,12 @@ class JUnitExecutors {
           }
           groupedTestResults[testRunInfo.className]?.add(testRunInfo)
```
phiSgr, line 30: Replace with `.groupBy`?

executors/src/main/kotlin/JUnitExecutors.kt
```diff
@@ -29,13 +29,12 @@ class JUnitExecutors {
...
         }
-        print(mapper.writeValueAsString(groupedTestResults))
+        standardOutput.print(mapper.writeValueAsString(groupedTestResults))
       }
       catch (e: Exception) {
-        System.setOut(standardOutput)
-        print("[\"")
+        standardOutput.print("[\"")
         e.printStackTrace()
```
phiSgr, line 36: It seems to me that it's writing to `PrintStream(errorStream)`. Then the BAOS is discarded. Should it be sent to stderr?

phiSgr:
> possible to add a test

Is the modified test sufficient?
https://github.com/JetBrains/kotlin-compiler-server/pull/618#discussion_r1148465908

phiSgr:
Following up on this.

I wasn't very clear in my previous comment. I have changed the test which now guards against the bug. I don't think a new additional test is needed.

The test that I modified, `base coroutines test 7` used to only test for substring match.
Now it is testing for full equality.

phiSgr:
This bug has confused another person.

https://stackoverflow.com/questions/77619172/why-does-the-kotlin-coroutine-output-a-b-even-with-a-delay

phiSgr:
I found another SO question caused by this bug.
Imagine how many people are confused by this but didn't ask a question.

https://stackoverflow.com/questions/71918773/why-does-coroutine-in-background-execute-before-the-main-thread-in-this-kotlin-c

nikpachoo:
@phiSgr, thank you for bringing this up! 
We are reviewing the PR, and I hope we can merge it this week.

executors/src/main/kotlin/JUnitExecutors.kt
nikpachoo: Thanks for the suggestion, replaced.

executors/src/main/kotlin/JUnitExecutors.kt
nikpachoo: Well, hard to say. The code in the `catch` block seems nonfunctional, regardless of the changes. I tried to reproduce a case when the execution flow reached this part of the code and got a serialization error.

nikpachoo (approved):

nikpachoo:
No objections from my side.
@phiSgr, thank you for the contribution!

phiSgr:
It seems to me that it's now deployed to play.kotlinlang.org only in version 2.0.0-RC2. Will it be back-ported to older versions as well?

nikpachoo:
@phiSgr, 
1.9.24 version released today contains the fix ✅ 
There are no plans to back-port it to all the previous versions; their usage is relatively low.
Thank you for your participation ❤️

Issue #579, opened 2022-12-05
Cannot inline bytecode built with JVM target 11 into bytecode that is being built with JVM target 1.8


sheu:
I cloned this project and added my own kotlin library that is compiled with java 11 as the target.  I however cannot run code that requires fold operation based on the result of library operations.  The server returns  this error:  "`Cannot inline bytecode built with JVM target 11 into bytecode that is being built with JVM target 1.8. Please specify proper '-jvm-target' option`"

What JVM target is used to the the code passed my `/run` endpoint?.  How can I specify the jvm target for run calls?

AlexanderPrendota:
I think it's duplicate. https://github.com/AlexanderPrendota/kotlin-compiler-server/issues/576

sheu:
I rebased and I still come across this issue. I also updated the source compatibility but I still get the same error. 
Also looking at #576 I cannot figure out what upgrade was done to resolve the issue.

My understanding is that when a call to /run is made, the compiler saver will compile and run the code.  I think this is where the issue is because the default target is 1.8 if not specified.  

I tried making this change: 
`private val additionalCompilerArguments: List<String> = listOf(
      "-opt-in=kotlin.ExperimentalStdlibApi",
      "-opt-in=kotlin.time.ExperimentalTime",
      "-opt-in=kotlin.RequiresOptIn",
      "-opt-in=kotlin.ExperimentalUnsignedTypes",
      "-opt-in=kotlin.contracts.ExperimentalContracts",
      "-opt-in=kotlin.experimental.ExperimentalTypeInference",
      "-jvm-target 11",
      "-Xcontext-receivers",
      "-XXLanguage:+RangeUntilOperator"
    )`
This did not help either. What am I missing ?

sheu:
After some debugging, I found out that the compiler configurations that is built in `component/KotlinEnvironment.kt` does not not include the jvm target and setting this in the `additionalCompilerArguments` does not help either.

The quick fix for me was to add the lines 
`put(JVMConfigurationKeys.JVM_TARGET, JvmTarget.JVM_11)` in the method `createConfiguration()`

PR #621, opened 2023-03-20
Use jvmtoolchain to setup java source compatibility

hfhbd:
This sets the correct JDK used by Gradle too. Otherwise, you need to change the JDK in IDEA project settings manually.

This would also allow you to remove the setup java steps in GitHub CI.


AlexanderPrendota:
Hi, I'm not 100% sure we need such changes here. What is the main problem you faced?

hfhbd:
My main problem was manually switching the Java SDK. With `jvmToolchain` this is done by Gradle.

Other problems: 
https://github.com/JetBrains/kotlin-compiler-server/issues/579

PR #633, opened 2023-05-04
KTL-1017 Compiler-Server: unify version definition

nikpachoo:
https://youtrack.jetbrains.com/issue/KTL-1017/Compiler-Server-unify-version-definition


docker-image-build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+kotlin_version=$(grep "systemProp.kotlinVersion" gradle.properties | cut -d'=' -f2)
```
zhelenskiy, line 3: `awk '{ if ($1=="systemProp.kotlinVersion") { print $2; } }' FS='=' ./gradle.properties` seems to be better solution.

zhelenskiy (changes requested):

docker-image-build.sh
nikpachoo: updated

zhelenskiy (approved):

gradle.properties
```diff
@@ -1,5 +1,6 @@
 systemProp.kotlinVersion=1.8.21
 systemProp.kotlinIdeVersion=1.8.21-release-380
+systemProp.kotlinIdeVersionSuffix=IJ4167.29
```
nikpachoo, line 3: @woainikk, JFI this version is now also in the configuration file.

woainikk (approved):

PR #634, opened 2023-05-09
chore(deps): Use 223 platform kotlin ide artifacts

woainikk:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openjdk:11.0.16-jdk as build
+FROM openjdk:17-oracle as build
```
nikpachoo, line 1: openjdk image is officially deprecated and all users are recommended to find and use suitable replacements ASAP.
Since we are updating the version, I suggest also change the vendor. Let's use [amazoncorretto](https://hub.docker.com/_/amazoncorretto).

nikpachoo:
⚠️ PLEASE, DON'T MERGE IT. 

The first step should be rising runtime version to 17 in AWS.
I will do it, and then will merge this PR.

nikpachoo (approved):

PR #638, opened 2023-05-21
KTL-1102 Keep the dependencies up-to-date with Dependabot

nikpachoo:
I suggest trying dependabot instead of renovate because of the easy integration with GitHub.


nikpachoo:
@AlexanderPrendota, I'd like to merge it to see how Dependabot works. Feel free to comment on post-review stage, I will take into account your objections.

PR #679, opened 2023-09-12
js legacy exception

ilgonmic:


src/main/kotlin/com/compiler/server/controllers/CompilerRestController.kt
```diff
@@ -1,5 +1,6 @@
 package com.compiler.server.controllers
 
+import com.compiler.server.exceptions.LegacyJsException
```
nikpachoo, line 3: I don't see where this import is used in the file. Should it be removed?

nikpachoo (approved):

PR #681, opened 2023-09-18
Use wasm frontend for wasm target

ilgonmic:


nikpachoo (approved):
@ilgonmic, please change target to the latest version (1.9.20-RC2)

PR #695, opened 2023-10-25
Add JVM bytecode generation for Kotlin/JVM

zhelenskiy:


src/main/kotlin/com/compiler/server/compiler/components/KotlinCompiler.kt
```diff
@@ -117,22 +138,23 @@ class KotlinCompiler(
...
-  ): ExecutionResult = try {
+    block: (output: OutputDirectory, compilation: JvmClasses) -> JvmExecutionResult
+  ): JvmExecutionResult = try {
     when (val compilationResult = compile(files)) {
       is Compiled<JvmClasses> -> {
         usingTempDirectory { outputDir ->
           val output = write(compilationResult.result, outputDir)
           block(output, compilationResult.result).also {
             it.addWarnings(compilationResult.compilerDiagnostics)
+            it.addByteCode(compilationResult.result)
```
nikpachoo, line 149: The changes affect the `execute` method, which is a default and often executes.
As I see it, the call `asHumanReadable` could be time-consuming since it performs bytecode parsing and text processing, which may not be trivial operations. Considering there is no need to show bytecode for each request, I suggest adding a separate method for it to run this code only when needed. 
It also makes the response at least 2x bigger, which is not optimal because we don't want to show it on Play now. It may be an option, but first, we must complete the current Survey on Play and make the UX adjustment based on it.
So, my propose, make a separate action to get byte code.

nikpachoo (changes requested):

src/main/kotlin/com/compiler/server/compiler/components/KotlinCompiler.kt
zhelenskiy: Such parsing is many times faster than the preceding compilation. The bytecode size itself is also many times smaller than the one from Wasm because it does not contain shrunken StdLib.

So, I find suggested complication redundant.

nikpachoo (approved):

PR #707, opened 2023-11-23
#2: Replace compiler internal API with compiler main function call besides completion

zoobestik:


nikpachoo:
#709 [709](https://github.com/JetBrains/kotlin-compiler-server/pull/709)

PR #713, opened 2023-12-28
Remove kotlin-ide artifacts, indexation, and turn off completion that used them

zhelenskiy:
Kotlin-ide artifacts are not published anymore since K2.


zhelenskiy:
@dkrasnoff

PR #719, opened 2024-01-30
Add support of compiler plugins for Wasm and prepare for JS

ilgonmic:
- extract runtime dependencies for compile process to separate submodule (`:dependencies`)
- support compiler plugins for JS and WASM, but only active for Wasm now
- Indexation process now consider compiler plugins when it enables (now for only Wasm)


nikpachoo (approved):

nikpachoo:
Tested in the prod environment, everything works fine.

PR #723, opened 2024-04-19
Initial support for Swift export

sbogolepov:
Some notes:
* We re-use Kotlin/Wasm stdlib klib to avoid downloading huge Kotlin/Native distribution. Should be fine as for now, Swift export supports a limited set of declarations from stdlib.
* Re-use existing endpoints for js/wasm translation.
* All Analysis API & Swift export dependencies are hidden in a separate module to avoid cluttering the main `build.gradle.kts`.

There might be bugs. We will squash them as soon as possible :)


nikpachoo (approved):

nikpachoo:
DO NOT MERGE IT. 
The Swift export works on RC2.

nikpachoo:
A note for the devs: in prod, we use "swift-export-v1" branch

swift-export-playground/src/main/kotlin/Runner.kt
```diff
@@ -0,0 +1,89 @@
...
+                    val parent = topLevelDeclaration.parent as? SirMutableDeclarationContainer
+                        ?: error("top level declaration can contain only module or extension to package as a parent")
+                    parent.addChild { topLevelDeclaration }
+                }
+            }
+        }
+        SirAsSwiftSourcesPrinter.print(
+            sirModule,
+            stableDeclarationsOrder = true,
+            renderDocComments = true,
```
mMaxy, line 44: This should be false - we do not support them currently, see KT-68744 for more information.

zhelenskiy:
Are there any updates?

nikpachoo:
@sbogolepov, could you please rebase you changes on the master?

sbogolepov:
Sure, done. Now that we have a proper publishing of Swift export artifacts the dependency hell is gone, and everything looks   more or less clean.

dkrasnoff:
The task https://youtrack.jetbrains.com/issue/KTL-2801 assumes closing this MR

PR #733, opened 2024-05-14
Upgrade to the latest kotlinx-datetime

dkhalanskyjb:
By the way, is there a reason `kotlinx-coroutines-test` is an old version?


nikpachoo (approved):

nikpachoo:
@dkhalanskyjb, thanks for the update.

> By the way, is there a reason kotlinx-coroutines-test is an old version?
Not really, it's a blind spot in the current release process. I think we should explicitly mention the playground in the libraries'  release checklists, so when a new version is released, we will also have an update on the playground.

PR #737, opened 2024-05-16
Add Fir extended checkers

zhelenskiy:


zhelenskiy:
I extracted updating .json data to a different commit to make the PR reviewable.

zhelenskiy:
I uncommented tests, but there is currently some compiler bug that needs being investigated first: `java.lang.IllegalArgumentException: source must not be null`

zhelenskiy:
I'll create a new review for 2.0.20-beta1, as a related compiler bug is fixed there.

PR #746, opened 2024-08-09
Added kotlinx-io support

dkrasnoff:
Added support for https://github.com/Kotlin/kotlinx-io library, but without file permissions

### DIFF
--- a/src/test/kotlin/com/compiler/server/IORunnerTest.kt
+++ b/src/test/kotlin/com/compiler/server/IORunnerTest.kt
@@ -1,0 +1,24 @@
+package com.compiler.server
+
+import com.compiler.server.base.BaseExecutorTest
+import org.junit.jupiter.api.Test
+
+class IORunnerTest : BaseExecutorTest() {
+
+    @Test
+    fun `kotlinx io basic test`() {
+        run(
+            code = """
+            import kotlinx.io.*
+            
+            fun main() {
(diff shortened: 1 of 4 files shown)


src/test/kotlin/com/compiler/server/IORunnerTest.kt
```diff
@@ -0,0 +1,24 @@
...
+    fun `kotlinx io basic test`() {
+        run(
+            code = """
+            import kotlinx.io.*
+            
+            fun main() {
+                val source = "Hello world".byteInputStream().asSource()
+                val buffered = source.buffered()
+                println(buffered.readString())
+
```
dkrasnoff, line 18: I'd like to highlight that using files will not be permitted. E.g. line like this:
```kotlin 
Path("Main.kt").source().readString()
```
will fail on server side because:
1) It uses `Main.kt` file and doesn't have file any `java.io.FilePermission`
2) It uses "java.io.tmpdir" in [kotlinx-io](https://github.com/Kotlin/kotlinx-io/blob/8a8d1d5b51dc4735b84ebd9b8641639bd80f88a4/core/jvm/src/files/FileSystemJvm.kt#L113), which is not covered with `permission java.util.PropertyPermission "kotlinx.io.*", "read";`

nikpachoo (approved):
Thank you!

PR #752, opened 2024-08-20
Added version catalog

dkrasnoff:
Added Version catalog https://docs.gradle.org/current/userguide/platforms.html#sub:central-declaration-of-dependencies and updated Gradle to 8.9


nikpachoo:
`coroutines-test test() FAILED`
Does it work locally?

dkrasnoff:
> `coroutines-test test() FAILED` Does it work locally?

Yeah, it works locally, I'm trying to understand what happens on CI

dkrasnoff:
@nikpachoo, I was wrong in the previous message. The test passed locally because of `indexes.json,` but it didn't change between changes. When I removed it manually, the problem was reproduced.
This problem is reproduced even in master since you've merged https://github.com/JetBrains/kotlin-compiler-server/pull/741 .
To make this pull request green, I've returned `kotlinx-coroutines-test` to `1.6.4` and `kotlinx-coroutines-core-jvm` to `1.7.3`, but generally, I don't understand the root cause well, because I'm not familiar enough with `indexation` module.

nikpachoo (approved):

PR #768, opened 2024-09-13
Provide the same java for child process in JavaExecutor as parent one

dkrasnoff:
In this pull request:

- Provided the same java for child process in JavaExecutor as parent one (e.g., on TeamCity, just running "java ..." leads to applying the wrong java). So we parse the Java home system property of the parent process to run the correct Java.
- Added [Gradle build scan ](https://docs.gradle.org/current/userguide/build_scans.html)to improve build info
- Set java 17 to [Gradle daemon](https://docs.gradle.org/current/userguide/gradle_daemon.html#daemon_jvm_criteria)


nikpachoo (approved):

CristianGM (commented):
Sorry, I forgot about it.

You may want to add buildCache

PR #822, opened 2025-06-18
KTL-2176: Change the logic of comparing compiled code to comparing ex…

dkrasnoff:
We had a problem that with each new version of Kotlin, the JS tests failed.
That happened because we compared translated JS code. With the current changes, we decided to compare not the translated code but the result of its execution.


gradle/libs.versions.toml
```diff
@@ -67,6 +68,8 @@ compose-material3 = { group = "org.jetbrains.compose.material3", name = "materia
 compose-components-resources = { group = "org.jetbrains.compose.components", name = "components-resources", version.ref = "compose" }
 kotlin-serialization-plugin = {group= "org.jetbrains.kotlin", name="kotlin-serialization-compiler-plugin", version.ref = "kotlin"}
 gradle-develocity = {group = "com.gradle", name= "develocity-gradle-plugin", version.ref = "gradle-develocity"}
+graalvm-polyglot = {group = "org.graalvm.polyglot", name= "polyglot", version.ref = "graalvm"}
```
dkrasnoff: I'd like to highlight that I'm using GraalVM to run JavaScript code in Kotlin tests, because I'm not very experienced in other possible solutions. But maybe you can suggest another approach.

ilgonmic:
We have following way to run JS scripts on top of Node.js
https://github.com/JetBrains/kotlin-compiler-server/blob/5a279510a983c292c893a99d20f8c2893d3e9bd1/src/test/kotlin/com/compiler/server/generator/TestProjectRunner.kt#L220-L223

gradle/libs.versions.toml
dkrasnoff: @ilgonmic, thanks for your suggestion. I've rewritten the code with your approach. Please check it.

ilgonmic (approved):

PR #878, opened 2025-10-01
KTL-3015 implement Kotlin LSP proxy for completions

S-furi:
This PR introduces a new way of performing completions requests leveraging [Kotlin-lsp](https://github.com/Kotlin/kotlin-lsp). Two approaches are included:

- **RESTful** on `api/compiler/lsp/complete`;
- **WebSocket** approach.

Currently completions with this approach supports only the latest Kotlin version, but future developments include the support of such feature by using a dedicated client and workspace for each Kotlin version supported by the kotlin-compiler-server.
Moreover, the support for different platforms than Kotlin/JVM is strictly bound to what Kotlin-lsp supports, namely only Kotlin/JVM projects as of today.

### DIFF
--- a/common/src/main/kotlin/model/Completion.kt
+++ b/common/src/main/kotlin/model/Completion.kt
@@ -34,7 +34,7 @@ data class ImportInfo(
 fun completionTextFromFullName(fullName: String): String {
   var completionText = fullName
   var position = completionText.indexOf('(')
-  if (position != -1) {
+  if (position != -1 && position > 0) {
     if (completionText[position - 1] == ' ') position -= 2
     if (completionText[position + 1] == ')') position++
     completionText = completionText.substring(0, position + 1)
--- a/src/main/kotlin/com/compiler/server/configuration/WebSocketConfig.kt
+++ b/src/main/kotlin/com/compiler/server/configuration/WebSocketConfig.kt
@@ -24,8 +24,6 @@ class WebSocketConfig : WebSocketConfigurer {
     @Bean
     fun createWebSocketContainer(): ServletServerContainerFactoryBean {
         return ServletServerContainerFactoryBean().apply {
-            maxTextMessageBufferSize = (1_000_000) // 1 MB
-            maxBinaryMessageBufferSize = (1_000_000)
             maxSessionIdleTimeout = 0L
         }
     }
--- a/src/main/kotlin/com/compiler/server/configuration/WebSocketConfig.kt
+++ b/src/main/kotlin/com/compiler/server/configuration/WebSocketConfig.kt
@@ -2,6 +2,8 @@ package com.compiler.server.configuration
 
 import com.compiler.server.controllers.LspCompletionWebSocketHandler
 import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
 import org.springframework.web.socket.config.annotation.EnableWebSocket
@@ -11,6 +13,8 @@ import org.springframework.web.socket.server.standard.ServletServerContainerFact
 
 @Configuration
 @EnableWebSocket
+@ConditionalOnWebApplication(type = ConditionalOnWebApplication.Type.SERVLET)
+@ConditionalOnProperty(name = ["websocket.enabled"], havingValue = "true")
 class WebSocketConfig : WebSocketConfigurer {
 
     @Autowired
--- a/src/main/kotlin/com/compiler/server/configuration/WebSocketConfig.kt
+++ b/src/main/kotlin/com/compiler/server/configuration/WebSocketConfig.kt
@@ -2,10 +2,12 @@ package com.compiler.server.configuration
 
 import com.compiler.server.controllers.LspCompletionWebSocketHandler
 import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
 import org.springframework.web.socket.config.annotation.EnableWebSocket
 import org.springframework.web.socket.config.annotation.WebSocketConfigurer
 import org.springframework.web.socket.config.annotation.WebSocketHandlerRegistry
+import org.springframework.web.socket.server.standard.ServletServerContainerFactoryBean
 
 @Configuration
 @EnableWebSocket
@@ -17,5 +19,14 @@ class WebSocketConfig : WebSocketConfigurer {
     override fun registerWebSocketHandlers(registry: WebSocketHandlerRegistry) {
         registry.addHandler(handler, "/lsp/complete")
             .setAllowedOrigins(ACCESS_CONTROL_ALLOW_ORIGIN_VALUE)
+    }
+
+    @Bean
+    fun createWebSocketContainer(): ServletServerContainerFactoryBean {
+        return ServletServerContainerFactoryBean().apply {
+            maxTextMessageBufferSize = (1_000_000) // 1 MB
+            maxBinaryMessageBufferSize = (1_000_000)
+            maxSessionIdleTimeout = 0L
+        }
     }
 }
--- a/src/main/kotlin/com/compiler/server/controllers/LspCompletionWebSocketHandler.kt
+++ b/src/main/kotlin/com/compiler/server/controllers/LspCompletionWebSocketHandler.kt
@@ -16,12 +16,12 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.channels.BufferOverflow
+import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableSharedFlow
-import kotlinx.coroutines.flow.collectLatest
+import kotlinx.coroutines.flow.conflate
+import kotlinx.coroutines.flow.consumeAsFlow
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.sync.Mutex
-import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withTimeoutOrNull
 import model.Completion
 import org.slf4j.LoggerFactory
@@ -38,32 +38,46 @@ import kotlin.time.Duration.Companion.seconds
 class LspCompletionWebSocketHandler(
     private val lspProxy: KotlinLspProxy,
     private val kotlinProjectExecutor: KotlinProjectExecutor,
-): TextWebSocketHandler() {
+) : TextWebSocketHandler() {
     private val job = SupervisorJob()
     private val scope = CoroutineScope(
         Dispatchers.IO + job + CoroutineName("LspCompletionWebSocketHandler")
     )
-
-    private val activeSession = ConcurrentHashMap<String, WebSocketSession>()
     private val logger = LoggerFactory.getLogger(LspCompletionWebSocketHandler::class.java)
 
-    private val sessionLocks = ConcurrentHashMap<String, Mutex>()
     private val completionsJob = ConcurrentHashMap<String, Job>()
-    private val sessionFlows = ConcurrentHashMap<String, MutableSharedFlow<CompletionRequest>>()
+    private val sessionMailbox = ConcurrentHashMap<String, Channel<CompletionRequest>>()
+
+    private val responseJobs = ConcurrentHashMap<String, Job>()
+    private val outgoingResponsesFlows = ConcurrentHashMap<String, MutableSharedFlow<Response>>()
 
     override fun handleTextMessage(session: WebSocketSession, message: TextMessage) {
         val request = session.decodeCompletionRequestFromTextMessage(message) ?: return
-        sessionFlows[session.id]?.tryEmit(request)
+        sessionMailbox[session.id]?.trySend(request)
     }
 
     override fun afterConnectionEstablished(session: WebSocketSession) {
-        activeSession[session.id] = session
+        val mailbox = Channel<CompletionRequest>(capacity = Channel.CONFLATED).also {
+            sessionMailbox[session.id] = it
+        }
 
-        val flow =  MutableSharedFlow<CompletionRequest>(
-            replay = 0,
-            extraBufferCapacity = 1,
+        val responseFlow = MutableSharedFlow<Response>(
+            extraBufferCapacity = 32,
             onBufferOverflow = BufferOverflow.DROP_OLDEST,
-        ).also { sessionFlows[session.id] = it }
+        ).also { outgoingResponsesFlows[session.id] = it }
+
+        val responseJob = scope.launch {
+            try {
+                responseFlow.collect { response ->
+                    if (session.isOpen) session.sendMessage(TextMessage(response.toJson()))
+                }
+            } catch (t: Throwable) {
+                if (t !is CancellationException) {
+                    logger.warn("Error collecting responses for client ${session.id}: ${t.message}")
+                }
+            }
+        }
+        responseJobs[session.id] = responseJob
 
         val completionWorker = scope.launch {
             with(session) {
@@ -78,38 +92,7 @@ class LspCompletionWebSocketHandler(
                 sendResponse(Response.Init(id))
             }
 
-            flow
-                .collectLatest { req ->
-                    val available = withTimeoutOrNull(LSP_TIMEOUT_WAIT_TIME) {
-                        while (!lspProxy.isAvailable()) {
-                            delay(LSP_TIMEOUT_POLL_INTERVAL)
-                        }
-                        true
-                    } ?: false
-
-                    if (!available) {
-                        session.sendResponse(Response.Error(message = "LSP not available", req.requestId))
-                        return@collectLatest
-                    }
-
-                    try {
-                        val res = kotlinProjectExecutor.completeWithLsp(
-                            clientId = session.id,
-                            project = req.project,
-                            line = req.line,
-                            character = req.ch
-                        )
-                        session.sendResponse(Response.Completions(completions = res, requestId = req.requestId))
-                    } catch (e: Exception) {
-                        logger.warn("Completion processing failed for client ${session.id}:", e)
-                        session.sendResponse(
-                            Response.Error(
-                                message = "Completion failed: ${e.message}",
-                                requestId = req.requestId
-                            )
-                        )
-                    }
-                }
+            mailbox.processIncomingRequests(session)
         }
         completionWorker.invokeOnCompletion { completionsJob.remove(session.id) }
         completionsJob[session.id] = completionWorker
@@ -124,27 +107,49 @@ class LspCompletionWebSocketHandler(
     override fun handleTransportError(session: WebSocketSession, exception: Throwable) {
         handleClientDisconnected(session.id)
         session.close(CloseStatus.SERVER_ERROR)
-        logger.debug("Lsp client transport error: {} ({}, {})", session.id, exception.cause, exception.message)
+        logger.info("Lsp client transport error: {} ({}, {})", session.id, exception.cause, exception.message)
     }
 
     private fun handleClientDisconnected(clientId: String) {
-        activeSession.remove(clientId)
-        sessionFlows.remove(clientId)
         completionsJob.remove(clientId)?.cancel()
+        responseJobs.remove(clientId)?.cancel()
+        sessionMailbox.remove(clientId)
+        outgoingResponsesFlows.remove(clientId)
         scope.launch { lspProxy.onClientDisconnected(clientId) }
     }
 
-    private suspend fun WebSocketSession.sendResponse(response: Response) {
-        val mutex = sessionLocks.computeIfAbsent(id) { Mutex() }
-        try {
-            mutex.withLock {
-                if (isOpen) sendMessage(TextMessage(response.toJson()))
+    private suspend fun Channel<CompletionRequest>.processIncomingRequests(session: WebSocketSession) {
+        consumeAsFlow().conflate().collect { req ->
+            val available = withTimeoutOrNull(LSP_TIMEOUT_WAIT_TIME) {
+                while (!lspProxy.isAvailable()) {
+                    delay(LSP_TIMEOUT_POLL_INTERVAL)
+                }
+                true
+            } ?: false
+
+            if (!available) {
+                session.sendResponse(Response.Error(message = "LSP not available", req.requestId))
+                return@collect
             }
-        } catch (e: Exception) {
-            if (e !is CancellationException) {
-                logger.warn("Error sending message to client $id:", e.message)
+
+            try {
+                val res = kotlinProjectExecutor.completeWithLsp(
+                    clientId = session.id, project = req.project, line = req.line, character = req.ch
+                )
+                session.sendResponse(Response.Completions(completions = res, requestId = req.requestId))
+            } catch (e: Exception) {
+                logger.warn("Completion processing failed for client ${session.id}:", e)
+                session.sendResponse(
+                    Response.Error(
+                        message = e.message ?: "Unknown error", requestId = req.requestId
+                    )
+                )
             }
         }
+    }
+
+    private fun WebSocketSession.sendResponse(response: Response) {
+        outgoingResponsesFlows[id]?.tryEmit(response)
     }
 
     private fun WebSocketSession.decodeCompletionRequestFromTextMessage(message: TextMessage): CompletionRequest? =
--- a/src/main/kotlin/com/compiler/server/controllers/LspCompletionWebSocketHandler.kt
+++ b/src/main/kotlin/com/compiler/server/controllers/LspCompletionWebSocketHandler.kt
@@ -56,7 +56,7 @@ class LspCompletionWebSocketHandler(
     override fun handleTextMessage(session: WebSocketSession, message: TextMessage) {
         lspProxy.requireAvailable()
         val request = session.decodeCompletionRequestFromTextMessage(message) ?: return
-        sessionFlows[session.id]?.tryEmit(request)
+        scope.launch { sessionFlows[session.id]?.emit(request) }
     }
 
     @OptIn(FlowPreview::class)
--- a/src/main/kotlin/com/compiler/server/controllers/LspCompletionWebSocketHandler.kt
+++ b/src/main/kotlin/com/compiler/server/controllers/LspCompletionWebSocketHandler.kt
@@ -18,10 +18,8 @@ import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.channels.getOrElse
-import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.withTimeoutOrNull
 import model.Completion
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Component
@@ -30,8 +28,6 @@ import org.springframework.web.socket.TextMessage
 import org.springframework.web.socket.WebSocketSession
 import org.springframework.web.socket.handler.TextWebSocketHandler
 import java.util.concurrent.ConcurrentHashMap
-import kotlin.time.Duration.Companion.milliseconds
-import kotlin.time.Duration.Companion.seconds
 
 @Component
 class LspCompletionWebSocketHandler(
@@ -127,17 +123,17 @@ class LspCompletionWebSocketHandler(
                 session.sendResponse(Response.Discarded(requestId = req.requestId))
                 req = next
             }
-            val available = withTimeoutOrNull(LSP_TIMEOUT_WAIT_TIME) {
-                while (!lspProxy.isAvailable()) {
-                    delay(LSP_TIMEOUT_POLL_INTERVAL)
-                }
-                true
-            } ?: false
 
-            if (!available) {
-                session.sendResponse(Response.Error(message = "LSP not available", req.requestId))
-                continue
-            }
+            runCatching { lspProxy.requireAvailable() }
+                .onFailure { throwable ->
+                    session.sendResponse(
+                        Response.Error(
+                            message = "LSP not available: ${throwable.cause}, ${throwable.message}",
+                            requestId = req.requestId,
+                        ),
+                    )
+                    continue
+                }
 
             try {
                 val res = kotlinProjectExecutor.completeWithLsp(
@@ -176,8 +172,6 @@ class LspCompletionWebSocketHandler(
 
     companion object {
         internal val objectMapper = ObjectMapper().apply { registerKotlinModule() }
-        private val LSP_TIMEOUT_WAIT_TIME = 10.seconds
-        private val LSP_TIMEOUT_POLL_INTERVAL = 100.milliseconds
     }
 }
 
--- a/src/main/kotlin/com/compiler/server/controllers/LspCompletionWebSocketHandler.kt
+++ b/src/main/kotlin/com/compiler/server/controllers/LspCompletionWebSocketHandler.kt
@@ -15,11 +15,8 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
-import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableSharedFlow
-import kotlinx.coroutines.flow.collectLatest
-import kotlinx.coroutines.flow.debounce
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
@@ -65,8 +62,8 @@ class LspCompletionWebSocketHandler(
         activeSession[session.id] = session
 
         val flow =  MutableSharedFlow<CompletionRequest>(
-            extraBufferCapacity = 1,
-            onBufferOverflow = BufferOverflow.DROP_OLDEST,
+//            extraBufferCapacity = 1,
+//            onBufferOverflow = BufferOverflow.DROP_OLDEST,
         ).also { sessionFlows[session.id] = it }
 
         val completionWorker = scope.launch {
@@ -82,8 +79,9 @@ class LspCompletionWebSocketHandler(
             }
 
             flow
-                .debounce { 200.milliseconds } // TODO: define a heuristic for average typing speed + frontend debounce
-                .collectLatest { req ->
+//                .debounce { 200.milliseconds } // TODO: define a heuristic for average typing speed + frontend debounce
+//                .collectLatest { req ->
+                .collect { req ->
                     val res = kotlinProjectExecutor.completeWithLsp(
                         clientId = session.id,
                         project = req.project,
--- a/src/main/kotlin/com/compiler/server/controllers/LspCompletionWebSocketHandler.kt
+++ b/src/main/kotlin/com/compiler/server/controllers/LspCompletionWebSocketHandler.kt
@@ -138,7 +138,7 @@ class LspCompletionWebSocketHandler(
                 }
             }
         } catch (e: Exception) {
-            logger.error("Error sending message to client $id:", e)
+            logger.warn("Error sending message to client $id:", e.message)
         }
     }
 
--- a/src/main/kotlin/com/compiler/server/controllers/LspCompletionWebSocketHandler.kt
+++ b/src/main/kotlin/com/compiler/server/controllers/LspCompletionWebSocketHandler.kt
@@ -15,6 +15,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.launch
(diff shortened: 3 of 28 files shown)


common/src/main/kotlin/model/Completion.kt
```diff
@@ -34,7 +34,7 @@ data class ImportInfo(
 fun completionTextFromFullName(fullName: String): String {
   var completionText = fullName
   var position = completionText.indexOf('(')
-  if (position != -1) {
+  if (position != -1 && position > 0) {
```
dkrasnoff: we can only leave `if (position > 0) {`

completions/src/test/resources/lsp/lsp-users-projects-root-test/settings.gradle.kts
```diff
@@ -0,0 +1 @@
+rootProject.name = "lsp-users-projects-root"
```
dkrasnoff, line 1: Why do we need this empty `lsp-users-projects-root` committed in this repo?

src/main/kotlin/com/compiler/server/configuration/WebSocketConfig.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.web.socket.config.annotation.EnableWebSocket
+import org.springframework.web.socket.config.annotation.WebSocketConfigurer
+import org.springframework.web.socket.config.annotation.WebSocketHandlerRegistry
+import org.springframework.web.socket.server.standard.ServletServerContainerFactoryBean
+
+@Configuration
+@EnableWebSocket
+@ConditionalOnWebApplication(type = ConditionalOnWebApplication.Type.SERVLET)
```
dkrasnoff: why do we need this annotation?

src/main/kotlin/com/compiler/server/configuration/WebSocketConfig.kt
```diff
@@ -0,0 +1,34 @@
...
+@ConditionalOnProperty(name = ["websocket.enabled"], havingValue = "true")
+class WebSocketConfig : WebSocketConfigurer {
+
+    @Autowired
+    private lateinit var handler: LspCompletionWebSocketHandler
```
dkrasnoff: nit: handler -> completionHandler

src/main/kotlin/com/compiler/server/configuration/WebSocketConfig.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    override fun registerWebSocketHandlers(registry: WebSocketHandlerRegistry) {
+        registry.addHandler(handler, "/lsp/complete")
+            .setAllowedOrigins(ACCESS_CONTROL_ALLOW_ORIGIN_VALUE)
+    }
+
+    @Bean
+    fun createWebSocketContainer(): ServletServerContainerFactoryBean {
```
dkrasnoff: why do we need this bean?

src/main/kotlin/com/compiler/server/controllers/CompilerRestController.kt
```diff
@@ -44,6 +44,13 @@ class CompilerRestController(private val kotlinProjectExecutor: KotlinProjectExe
     @RequestParam ch: Int
   ) = kotlinProjectExecutor.complete(project, line, ch)
 
+  @PostMapping("/lsp/complete")
+  suspend fun getKotlinLspCompleteEndpoint(
```
dkrasnoff: Let's move the whole completion logic to a separate Spring Boot application

src/main/kotlin/com/compiler/server/controllers/LspCompletionWebSocketHandler.kt
```diff
@@ -0,0 +1,194 @@
...
+                }
+            } catch (t: Throwable) {
+                if (t !is CancellationException) {
+                    logger.warn("Error collecting responses for client ${session.id}: ${t.message}")
+                }
+            }
+        }
+        responseJobs[session.id] = responseJob
+
+        val completionWorker = scope.launch {
```
dkrasnoff: We can extract lines from 77 to 93 to a separate method called `setUpCompletionWorker` (you can think about a better name)

completions/src/test/resources/lsp/lsp-users-projects-root-test/settings.gradle.kts
S-furi: This is the base project from which LSP will retrieve base libraries/sdks/project sources. As of now, this project just sets up a simple kotlin project with version `2.2.20`, but it is thanks to this base project that we could define libraries available in playground (e.g. `kotlinx.coroutines`) or common configurations. 
The LSP needs a gradle project in its local filesystem, and I though it could be handy to have the base project here in the compiler server repo in order to reflect the current completion's approach.

src/main/kotlin/com/compiler/server/controllers/LspCompletionWebSocketHandler.kt
```diff
@@ -0,0 +1,194 @@
...
+        val request = session.decodeCompletionRequestFromTextMessage(message) ?: return
+        incomingCompletionsRequests[session.id]?.trySend(request)
+    }
+
+    override fun afterConnectionEstablished(session: WebSocketSession) {
+        val incoming = Channel<CompletionRequest>(capacity = Channel.UNLIMITED).also {
+            incomingCompletionsRequests[session.id] = it
+        }
+
+        val responseFlow = MutableSharedFlow<Response>(
```
dkrasnoff: We can extract lines from 59 to 75 to a separate method called `setupResponseFlow` (you can think about a better name)

dkrasnoff:
Please add kdocs for all public methods and classes

completions/src/main/kotlin/completions/lsp/KotlinLspProxy.kt
```diff
@@ -0,0 +1,265 @@
...
+        val project = clientsProjects[clientId] ?: return emptyList()
+        val lspProject = lspProjects[project] ?: return emptyList()
+        val newContent = newProject.files.first().text
+        val documentToChange = project.files.first().name
+        changeDocumentContent(lspProject, documentToChange, newContent)
+        return getCompletions(lspProject, line, ch, documentToChange)
+    }
+
+    fun KotlinLspProxy.onClientConnected(clientId: String) {
+        val project = Project(files = listOf(ProjectFile(name = "$clientId.kt")))
```
dkrasnoff: I'd say each project should have `clientId`, but files can then be called `file1-$clientId.kt`, `file2-$clientId.kt`, etc, so that we will have a possibility to work with mutliple files

completions/src/main/kotlin/completions/lsp/KotlinLspProxy.kt
```diff
@@ -0,0 +1,265 @@
...
+            .also { clientsProjects[clientId] = it }
+        val lspProject = LspProject.fromProject(project).also { lspProjects[project] = it }
```
dkrasnoff: Not clear for me, why do we have project and lspProject, can it be merged into one entity here?

completions/src/main/kotlin/completions/lsp/KotlinLspProxy.kt
dkrasnoff: Is it hard to rewrite it to list of files instead of one file now?

src/main/kotlin/com/compiler/server/configuration/WebSocketConfig.kt
S-furi: Adding this helped to avoid an unwanted behaviour that I was encountering during testing. Having WebSocket that uses WebFlux for some of its components (current implementation still uses Servlet and could actually benefit from a full switch to WebFlux) made every test that used Servlet to load WS configuration. The problem is that in testing when using `@SpringBootTest` a lightweight Servlet instance will be brought up, which by default lacks WS support in a non-ws test, causing practically every test in this codebase fail for this reason. With this annotation the WS configuration is loaded  only with WebSocket tests then we enable it with the [right property](https://github.com/JetBrains/kotlin-compiler-server/blob/adece876b8ae75f9c9b680f817baad055e4b416b/src/test/kotlin/com/compiler/server/lsp/ws/KotlinLspProxyWSTest.kt#L32).

src/main/kotlin/com/compiler/server/controllers/LspCompletionWebSocketHandler.kt
```diff
@@ -0,0 +1,194 @@
...
+            with(session) {
+                logger.info("Lsp client connected: $id")
+                runCatching {
+                    lspProxy.requireAvailable()
+                }.onFailure {
+                    sendResponse(Response.Error("Proxy is still not connected to Language server"))
+                    return@launch
+                }
+                lspProxy.onClientConnected(id)
+                sendResponse(Response.Init(id))
```
dkrasnoff: This method looks inside like this `outgoingResponsesFlows[id]?.tryEmit(response)`, so when `outgoingResponsesFlows[id]` is null, we don't send a message, but continue to create a worker here. Is it right, or should we `return@launch` instead?

src/main/kotlin/com/compiler/server/configuration/WebSocketConfig.kt
S-furi: It's there so that it is possible to customise the underlying jakarta websocket's server container. The custom configuration should help to avoid closing WS sessions after periods of inactivity (typically 60s).

src/main/kotlin/com/compiler/server/controllers/LspCompletionWebSocketHandler.kt
```diff
@@ -0,0 +1,194 @@
...
+    private fun handleClientDisconnected(clientId: String) {
+        completionJobs.remove(clientId)?.cancel()
+        responseJobs.remove(clientId)?.cancel()
+        incomingCompletionsRequests.remove(clientId)
+        outgoingResponsesFlows.remove(clientId)
+        scope.launch { lspProxy.onClientDisconnected(clientId) }
+    }
+
+    private suspend fun Channel<CompletionRequest>.processIncomingRequests(session: WebSocketSession) {
+        while (true) {
```
dkrasnoff: do we really need an infinite loop here?

src/main/kotlin/com/compiler/server/controllers/LspCompletionWebSocketHandler.kt
```diff
@@ -0,0 +1,194 @@
...
+            val first = receiveCatching().getOrElse { return }
+            var req = first
+
+            while (true) {
+                val next = tryReceive().getOrNull() ?: break
+                session.sendResponse(Response.Discarded(requestId = req.requestId))
+                req = next
+            }
+
+            runCatching { lspProxy.requireAvailable() }
```
dkrasnoff: this logic can be extracted to a private method and reused in other places

completions/src/test/resources/lsp/lsp-users-projects-root-test/settings.gradle.kts
dkrasnoff: Hm, then it looks more like a folder in `resources`

src/main/kotlin/com/compiler/server/configuration/WebSocketConfig.kt
dkrasnoff: Am I right, that extracting the whole completion to the separate subproject will allow you to make it WebFlux only and also get rid of these problems?

src/main/kotlin/com/compiler/server/configuration/WebSocketConfig.kt
dkrasnoff: Let's leave a comprehensive comment here then.

Also, am I right, that we can use 60000ms here instead of 0? If yes, then it is better to extract this value to application.yaml (application.properties) file and use the parameter everywhere it is needed

src/main/kotlin/com/compiler/server/configuration/WebSocketConfig.kt
S-furi: Yes it should really help, especially because the coexistence of Servlet and WebFlux in the same application require some tricks. I'm working on separating the completion logic to another subproject, then I'll try to migrate to WebFlux only.

src/main/kotlin/com/compiler/server/configuration/WebSocketConfig.kt
S-furi: Here we're using [0 to *disable* idle timeout](https://docs.oracle.com/javaee/7/api/javax/websocket/Session.html#setMaxIdleTimeout-long-). If you think that I should add it to application.yaml anyway just let me know.

completions/src/main/kotlin/completions/lsp/KotlinLspProxy.kt
S-furi: It shouldn't introduce big problems, but right now the LSP logic is implemented considering only a single file, meaning that every file is treated in isolation regardless of the same user. Unfortunately it is not trivial to make the user's files see each other with the current `isolated documents` LSP mode.

completions/src/main/kotlin/completions/lsp/KotlinLspProxy.kt
S-furi: They can be merged, but I needed a way to keep track of files in the file system, retrieve their lsp-compliant URIs and lsp documents versions. I though it was better to keep them separate being `Project` a simpler entity and already containing what's needed for most compiler server tasks.

src/main/kotlin/com/compiler/server/controllers/LspCompletionWebSocketHandler.kt
S-furi: we could actually put a condition on channel's state, like `while (!isClosedForReceive)` (*delicate coroutines api*) that should even make this function purpose clearer.

dkrasnoff: Yeap, sounds better

completions/src/main/kotlin/completions/lsp/KotlinLspProxy.kt
dkrasnoff: I expect that after extracting  completion to a separate subproject, you will not have a Project entity, so that you can only leave a LspProject

completions/src/main/kotlin/completions/lsp/KotlinLspProxy.kt
dkrasnoff: Got it, thanks!
Anyway, let's make multiple files here and pass to lsp only the first one. And leave a comment or todo, with the place where we choose only one file, with the explanation about LSP mode. 
Note: if it is a todo, then it should have the number of the task connected to it. If we don't have such a task, then you should create it.

src/main/kotlin/com/compiler/server/configuration/WebSocketConfig.kt
dkrasnoff: Yes, but should we disable the idle timeout or extend it by at least 60 seconds?

S-furi:
@dkrasnoff With recent commits I've made some changes that have probably changed the references to the request changes above. In short, I've
- Resolved conflicts `kotlin-community/dev` base branch
- Moved completion logic into a new springboot application module
- LSP-based completions now leverage `WebFlux`
  - Both REST endpoint (marked as suspendable) and WebSocket Implementation.
- Moved LSP related files from project root to its corresponding resources folder ([conversation reference](https://github.com/JetBrains/kotlin-compiler-server/pull/878#discussion_r2410101695))
- Applied suggested refactors

You might want to mark some conversations as resolved and maybe double-check the new WebSocket component for the refactoring suggestions you've made to the previous implementation.

If you want I can ping you when I think I've resolved everything, so that you just check once without having to check the code again.

completions/src/main/kotlin/completions/lsp/KotlinLspProxy.kt
S-furi: Even after extracting the completion logic to a separate subproject, the dependency on `Project` is still present due to the API contract that we have with frontend. Frontend will send a `Project` instance to completion endpoints (both REST and WS as of now), so we still need a way to map the incoming project plus the convenient `LspProject`. If the subproject will be lsp-based only I can merge the two entities, while I suggest to keep them separated in case it must be shared by other completion logics. In the meantime I'll take care in documenting `LspProject` so that its purpose is clear.

completions/src/main/kotlin/completions/lsp/KotlinLspProxy.kt
S-furi: I've added support for multiple documents, let me know if it's ok. 
As a sidenote, my main doubt is how we handle which document among the project's documents has to be picked for providing completions. Right now in the `KotlinLspProxy` I've changed completion functions signatures in order to accept a `ProjectFile` (which by default is the first file of the project), which now are unused due to the API format, i.e frontend just provide project + line + character. Once multiple files will be supported, the APIs should change in order to include such logic.

completions/src/test/kotlin/AbstractCompletionTest.kt
```diff
@@ -0,0 +1,95 @@
+import org.junit.jupiter.api.Test
+
+interface AbstractCompletionTest {
```
dkrasnoff: Remove `CompletionTest`, `ImportTest` and completion connected tests from `ConcurrencyRunnerTest` and move deleted tests to the new completions module, please

completions/src/test/resources/lsp/lsp-users-projects-root-test/settings.gradle.kts
dkrasnoff: Can we also remove `gradle`,`.gitattributes`, and `.gitignore` from this test project?

completions/build.gradle.kts
```diff
@@ -0,0 +1,27 @@
+
+plugins {
+    id("base-kotlin-jvm-conventions")
+    alias(libs.plugins.spring.dependency.management)
+    alias(libs.plugins.spring.boot)
+    alias(libs.plugins.kotlin.plugin.spring)
+}
+
+dependencies {
+    implementation("org.springframework.boot:spring-boot-starter-webflux")
```
dkrasnoff: lets put in `libs.versions.toml` as well

completions/build.gradle.kts
```diff
@@ -0,0 +1,27 @@
...
+    implementation(libs.org.eclipse.lsp4j)
+    implementation(libs.kotlinx.coroutines.reactor)
+    implementation(libs.kotlinx.serialization.core.jvm)
+    implementation(libs.kotlinx.serialization.json.jvm)
+    implementation(project(":common", configuration = "default"))
+
+    testImplementation(libs.kotlin.test)
+    testImplementation(libs.bundles.testcontainers)
+    testImplementation(libs.rector.test)
+    testImplementation("org.springframework.boot:spring-boot-starter-test") {
```
dkrasnoff: lets put in `libs.versions.toml` as well

completions/src/main/kotlin/completions/controllers/ws/LspCompletionWebSocketHandler.kt
```diff
@@ -0,0 +1,161 @@
...
+    private fun getCompletionRequests(session: WebSocketSession, sideSink: Sinks.Many<Response>): Flux<CompletionRequest> =
+        session.receive()
+            .map { it.payloadAsText }
+            .onBackpressureDrop { dropped ->
+                runCatching { objectReader.readValue<CompletionRequest>(dropped) }.onSuccess {
+                    sideSink.tryEmitNext(Response.Discarded(it.requestId))
+                }
+            }
+            .flatMap({ payload ->
+                val req = runCatching { objectReader.readValue<CompletionRequest>(payload) }.getOrNull()
```
dkrasnoff: Why do we need `objectReader`? Not clear for me, why just not use: `objectMapper.readValue(payload, CompletionRequest::class.java)`

completions/src/main/kotlin/completions/controllers/ws/LspCompletionWebSocketHandler.kt
```diff
@@ -0,0 +1,161 @@
...
+}
+
+private data class CompletionRequest(
+    val requestId: String,
+    val project: Project,
+    val line: Int,
+    val ch: Int,
+)
+
+private val objectMapper = ObjectMapper().apply { registerKotlinModule() }
```
dkrasnoff: lets set up it in Configuration calss, and use everywhere as a bean

completions/src/main/kotlin/completions/lsp/LspAvailabilityAdvice.kt
```diff
@@ -0,0 +1,14 @@
+package completions.lsp
+
+import completions.exceptions.LspUnavailableException
+import org.springframework.http.HttpStatus
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.ControllerAdvice
+import org.springframework.web.bind.annotation.ExceptionHandler
+
+@ControllerAdvice
+class LspAvailabilityAdvice {
```
dkrasnoff: Let's maybe move the controller package and write a class global exception handler for all exceptions, where we can handle all custom exceptions like this.

completions/src/main/kotlin/completions/lsp/LspCompletionParser.kt
```diff
@@ -0,0 +1,168 @@
...
+    private fun CompletionItem.sortingKey(): String = this.filterText ?: this.label
+
+    /**
+     * Ranking completions inspired by how VS-code does it. Firstly a simple
+     * fuzzy scoring is performed on what has been typed by the user so far,
+     * then we use [CompletionItem.sortText] to break ties.
+     *
+     * @param query the query the user has typed so far
+     */
+    fun List<CompletionItem>.rankCompletions(query: String): List<CompletionItem> =
```
dkrasnoff: nit: I used to conventions, where all public methods go on top and private go down

completions/src/main/kotlin/completions/lsp/LspCompletionParser.kt
```diff
@@ -0,0 +1,168 @@
...
+
+    /**
+     * Extracts the prefix of this [CompletionItem] that triggered the completion.
+     */
+    val CompletionItem.completionQuery: String?
+        get() = objectMapper.parseToJsonElement(data.toString()).jsonObject["additionalData"]
+                ?.jsonObject?.get("prefix")
+                ?.jsonPrimitive?.content
+
+    private fun fuzzyScore(query: String, candidate: String): Int {
```
dkrasnoff: Was it taken from VS code plugin?

completions/src/main/kotlin/completions/model/Project.kt
```diff
@@ -0,0 +1,26 @@
+package completions.model
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.fasterxml.jackson.annotation.JsonValue
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+data class Project(
```
dkrasnoff: lets move it to package completions.dto.api call it CompletionsRequest and remove `args` and `confType` as we don't need this fields

completions/src/main/kotlin/completions/controllers/ws/LspCompletionWebSocketHandler.kt
S-furi: object reader is an instance of object writer that is used only for instances of `CompletionRequest::class.java`, which in theory should improve a little bit deserialization performance. Even if we have to explicitly mark the generic parameter in `readValue<CompletionRequest>`, this is just for enabling type inference, that object reader can be used only for parsing `CompletionRequest`.
If it's confusing I think that I can remove it without any noticeable performance loss.

completions/src/main/kotlin/completions/lsp/LspCompletionParser.kt
S-furi: It's inspired and a simplification of how Monaco editor (the actual editor inside VS-Code) [performs completion matching](https://microsoft.github.io/monaco-editor/typedoc/interfaces/editor.ISuggestOptions.html). With LSP it's largely up to the editor to determine completion sorting, LSP just provides a `filterText: int` field that rarely yields good completion results on top.

completions/src/test/kotlin/ImportTest.kt
```diff
@@ -9,14 +7,14 @@ import kotlin.test.assertEquals
 
 // TODO(Dmitrii Krasnov): this test is disabled until KTL-2807 is fixed
```
dkrasnoff: I thought we could remove this todo and ignore the annotation here and in the concurrency test

completions/build.gradle.kts
```diff
@@ -0,0 +1,27 @@
+
+plugins {
+    id("base-kotlin-jvm-conventions")
+    alias(libs.plugins.spring.dependency.management)
```
dkrasnoff: What do you think about extracting base spring related things to Grdle convention plugin like `base-kotlin-jvm-conventions`, so we can share the same spring build setting across different modules.

completions/src/main/kotlin/completions/controllers/ws/LspCompletionWebSocketHandler.kt
dkrasnoff: No, thanks for the explanation.

completions/src/test/kotlin/ImportTest.kt
S-furi: Yes we can, but I wanted to discuss with you about a little detail: having developed basically two ways of performing completion (REST and WebSocket), we could take inspiration from current's [`CompletionTest`](https://github.com/S-furi/kotlin-compiler-server/blob/exp/isolated-documents/completions/src/test/kotlin/CompletionTest.kt), i.e. an abstract test that is reified for both implementations. In this way we reuse current tests, testing both completions approaches adding the least amount of code as possible. 

I was trying to implement this with `ImportTest`, but apparently LSP completions differs a little bit from what is expected in that test (e.g. old completions: `Random  (kotlin.random.Random)`, LSP-based completions: `Random (kotlin.random)`). In these cases, should I change the tests or should I adapt LSP-based completions to make those tests pass? (it is mostly regex and string manipulation).

completions/build.gradle.kts
S-furi: That's a great idea! I'm going to try and see what I can do after fixing requested changes!

completions/src/test/kotlin/ImportTest.kt
dkrasnoff: Yeap, you can manipulate with test data if it doesn't affect the main logic of the check like in your example.

completions/src/test/kotlin/ImportTest.kt
```diff
@@ -0,0 +1,213 @@
...
+
+    private fun doOpenBracketAfterImportCompletionTest(isJs: Boolean = false) {
+        val code = """
+            fun main() {
+                val s = sin$CARET_MARKER
+            }
+        """.trimIndent()
+
+        val foundCompletionsTexts = getCompletions(codeWithCaret = code, isJs = isJs).map { it.text }
+        val completions = listOf("kotlin.math.sin(")
```
S-furi: @dkrasnoff I think I'm not getting what is the expected behaviour here, and also in current version of playground. Even when looking at the HTTP responses of playground if I insert this code snippet, I'm receiving the following completion:
```json
{
	"1": {
		"text": "kotlin.math.sin(",
		"displayText": "sin(x: Double)  (kotlin.math.sin)",
		"tail": "Double",
		"import": "kotlin.math.sin",
		"icon": "method",
		"hasOtherImports": true
	}
}
```
`hasOtherImports` is marked as `true`, so the `text` field is replaced from `sin(` to the fully qualified name `kotlin.math.sin(`. However, I think this logic is wrong, because `hasOtherImports` should not be `true` due to the fact that the there aren't actually other imports that matches the same function signature, therefore `text` should be `sin(`. 

On the other hand, if this is a design choice (i.e. prefer fully qualified names over auto-import), please let me know.
I think the best way to understand how I am figuring out this logic can be found in [how I cleanup completions to `Completion`](https://github.com/JetBrains/kotlin-compiler-server/blob/61d399e3dce389b06719bdeeb024ea1ef0469124/completions/src/main/kotlin/completions/service/lsp/LspCompletionProvider.kt#L76) before sending them to clients.

completions/build.gradle.kts
S-furi: @dkrasnoff I've tried to create a common [`base-spring-boot-conventions`](https://github.com/JetBrains/kotlin-compiler-server/blob/7129757437bc162c06a61edb209f8a320c998df8/buildSrc/src/main/kotlin/base-spring-boot-conventions.gradle.kts), and I wanted to leverage as much as I could the version catalog file. As you already know, using it in the `buildSrc` involves some hacky solutions, and I've linked them above the code snippets. It works for the `:completions` subproject, however it fails in the base compiler application project. I've tried to add the convention plugin in the plugin definition but I'm getting the following error:

```
An exception occurred applying plugin request [id: 'base-spring-boot-conventions']
> Failed to apply plugin 'base-spring-boot-conventions'.
   > Plugin with id 'org.springframework.boot' not found.
```

Have you already encountered such problem? Do you have any idea on why this happens and how to solve it?

completions/src/test/kotlin/ImportTest.kt
dkrasnoff: It looks like we've discussed it on 1-1 and decided to add `import kotlin.math.sin`

buildSrc/src/main/kotlin/base-spring-boot-conventions.gradle.kts
```diff
@@ -0,0 +1,27 @@
+import org.gradle.accessors.dm.LibrariesForLibs
+import org.gradle.kotlin.dsl.the
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+    id("base-kotlin-jvm-conventions")
+}
+
+// https://github.com/gradle/gradle/issues/15383#issuecomment-779893192
```
dkrasnoff: good catch 👍  but what do you thing about using https://github.com/radoslaw-panuszewski/typesafe-conventions-gradle-plugin ?

completions/build.gradle.kts
dkrasnoff: lets try https://github.com/radoslaw-panuszewski/typesafe-conventions-gradle-plugin  and declare plugins via alias

completions/src/main/kotlin/completions/dto/api/Completion.kt
```diff
@@ -0,0 +1,44 @@
+package completions.dto.api
+
+import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.annotation.JsonProperty
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+data class Completion(
```
dkrasnoff: nit: Completion -> CompletionResponse

buildSrc/src/main/kotlin/base-spring-boot-conventions.gradle.kts
S-furi: It worked like a charm, thanks a lot for pointing that plugin out!

completions/build.gradle.kts
S-furi: that plugin fixed this issue!

completions/src/main/kotlin/completions/lsp/LspCompletionParser.kt
```diff
@@ -0,0 +1,142 @@
...
+    private fun CompletionItem.hasToBeImported(): Boolean {
+        val lookupObject = objectMapper.parseToJsonElement(data.toString())
+            .jsonObject["additionalData"]
+            ?.jsonObject?.get("model")
+            ?.jsonObject?.get("delegate")
+            ?.jsonObject?.get("delegate")
+            ?.jsonObject?.get("lookupObject")
+            ?.jsonObject?.get("lookupObject")
+
+        val importingStrategy =
```
dkrasnoff: val importingStrategy = lookupObject?.jsonObject?.get("options") ?: lookupObject

completions/src/main/kotlin/completions/lsp/LspCompletionParser.kt
```diff
@@ -0,0 +1,142 @@
...
+            if (lookupObject?.jsonObject?.get("options") != null) {
+                lookupObject.jsonObject["options"]
+            } else lookupObject
+
+        lookupObject?.jsonObject?.get("kind")?.let {
```
dkrasnoff: do it before getting importingStrategy

completions/src/main/kotlin/completions/lsp/LspCompletionParser.kt
```diff
@@ -0,0 +1,142 @@
...
+        "kotlin.jvm.internal",
+        "kotlin.coroutines.experimental.intrinsics",
+        "kotlin.coroutines.intrinsics",
+        "kotlin.coroutines.experimental.jvm.internal",
+        "kotlin.coroutines.jvm.internal",
+        "kotlin.reflect.jvm.internal",
+    )
+}
+
+private val objectMapper = Json { ignoreUnknownKeys = true }
```
dkrasnoff: Let's make it via bean

completions/src/main/kotlin/completions/configuration/lsp/LspProperties.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.stereotype.Component
+
+/**
+ * Simple LSP configuration properties which expose a way to be accessed
+ * even from non-spring managed components, i.e. [completions.lsp.client.LspConnectionManager].
+ */
+@Component
+@ConfigurationProperties(prefix = "lsp")
+data class LspProperties(
```
dkrasnoff, line 10: It can be simplified to just:
```
@ConfigurationProperties(prefix = "lsp")
data class LspProperties(
    val host: String,
    val port: Int,
    val reconnectionRetries: Int
)
```
and adding `@EnableConfigurationProperties(value = [LspProperties::class])` to main class

completions/src/main/kotlin/completions/configuration/JacksonConfig.kt
```diff
@@ -0,0 +1,19 @@
...
+
+@Configuration
+class JacksonConfig {
+
+    @Bean
+    fun objectMapper(): ObjectMapper =
+        ObjectMapper().registerModule(kotlinModule())
+
+    companion object {
+        fun JsonNode.walk(fieldName: String): JsonNode? = path(fieldName).takeIf { !it.isMissingNode }
```
dkrasnoff: it is not a configuration, but should be an util

completions/src/main/kotlin/completions/dto/api/CompletionResponse.kt
```diff
@@ -0,0 +1,44 @@
...
+            }
+            position = completionText.indexOf(":")
+            if (position != -1) completionText = completionText.take(position - 1)
+            return completionText
+        }
+    }
+}
+
+
+enum class Icon {
```
dkrasnoff: lets keep all enums in `enums` package

completions/src/main/kotlin/completions/lsp/LspCompletionParser.kt
```diff
@@ -0,0 +1,143 @@
...
+            .walk("additionalData")
+            ?.walk("model")
+            ?.walk("delegate")
+            ?.walk("delegate")
+            ?.walk("lookupObject")
+            ?.walk("lookupObject") ?: return true
+
+        val importingStrategy = lookupObject.walk("options") ?: lookupObject
+
+        lookupObject.walk("kind")?.let {
```
dkrasnoff, line 116: lets do it before:
```val importingStrategy = lookupObject.walk("options") ?: lookupObject```

File: completions/src/main/resources/lsp-users-projects-root/build.gradle.kts [hunk 35]
```diff
@@ -0,0 +1,9 @@
+package `lsp-users-projects-root`
```
dkrasnoff: can we remove it?

completions/build.gradle.kts
```diff
@@ -0,0 +1,24 @@
...
+    implementation(libs.org.eclipse.lsp4j)
+    implementation(libs.kotlinx.coroutines.reactor)
+    testImplementation(libs.kotlin.test)
+    testImplementation(libs.bundles.testcontainers)
+    testImplementation(libs.rector.test)
+}
+
+tasks.named<BootBuildImage>("bootBuildImage") {
+    val baseImageName = "sfuri/kotlin-compiler-server-completions-lsp"
+    // TODO: push docker image to JB registry
```
dkrasnoff: create a task for it please and link it here

completions/compose.yml
```diff
@@ -0,0 +1,21 @@
+services:
```
dkrasnoff, line 1: It will be great if you also add a dependency `org.springframework.boot:spring-boot-docker-compose`, and with `local` spring profile, it will use all the necessary Docker containers

completions/compose.yml
```diff
@@ -0,0 +1,21 @@
...
+  kotlin-lsp:
```
dkrasnoff, line 2: it is also make sense to add port forwanding like:
```
  kotlin-lsp:
    image: sfuri/intellij-lsp
    ports:
      - "9999:9999"
```

completions/src/main/resources/lsp-users-projects-root/build.gradle.kts
S-furi: Sure, my bad, it's automatically added when moving the folder around in IntelliJ. Fixed.

completions/src/main/kotlin/completions/configuration/lsp/LspProperties.kt
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Simple LSP configuration properties which expose a way to be accessed
+ * even from non-spring managed components, i.e. [completions.lsp.client.LspConnectionManager].
+ */
+@ConfigurationProperties(prefix = "lsp")
+data class LspProperties(
+    val host: String,
+    val port: Int,
+    val reconnectionRetries: Int,
+    val remoteWorkspaceRoot: String = "/lsp/workspaces/lsp-users-projects-root",
```
dkrasnoff, line 14: default value here and in application.yml are different, lets maybe remove default values from here?

completions/src/main/kotlin/completions/util/serializationUtil.kt
```diff
@@ -0,0 +1,7 @@
+package completions.util
+
+import com.fasterxml.jackson.databind.JsonNode
+
+object SerializationUtil {
```
dkrasnoff: the name of file starts from the lower letter

completions/src/main/kotlin/completions/configuration/lsp/LspProperties.kt
S-furi: Yes they're different because the `application*.yml` mirrors the remote workspace when using the docker container for `intellij-lsp`. The default values in `LspProperties` are the values to be used when running the LSP on the same machine as the completion application (or more in general if they refer to the same filesystem).

If you want I can keep the `application-local.yml` as is, due to the fact that it uses the dockerized version of the LSP through `spring-boot-docker-compose`, and maybe remove the default values on `application.yml` as it will be used in production and we don't know yet which configuration it'll have.

completions/src/main/resources/application-local.yml
```diff
@@ -0,0 +1,15 @@
+server:
+  port: ${COMPLETION_SERVICE_PORT:8082}
+lsp:
+  host: ${LSP_HOST:localhost}
+  port: ${LSP_PORT:9999}
+  reconnection-retries: 10
+  remote-workspace-root: /workspaces/lsp-users-projects-root
+spring:
+  docker:
+    compose:
```
dkrasnoff, line 10: lifecycle-management: start_and_stop
      enabled: true

dkrasnoff (approved):

PR #898, opened 2025-11-11
Adopt Compose wasm multi module compilation.

ilgonmic:
There is new compilation model in wasm with per-klib binary compilation


cache-maker/src/wasmJsMain/kotlin/main.kt
```diff
@@ -0,0 +1,65 @@
...
+import androidx.compose.material.Text
+import androidx.compose.runtime.*
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.ExperimentalComposeUiApi
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.window.ComposeViewport
+import kotlinx.browser.document
+
+@OptIn(ExperimentalComposeUiApi::class)
+fun main() {
```
dkrasnoff, line 1: why do we need this file?

common/src/main/kotlin/com/compiler/server/common/components/CliUtils.kt
```diff
@@ -0,0 +1,25 @@
+package com.compiler.server.common.components
+
+import java.io.File
+import java.nio.file.Path
+import java.util.*
+import kotlin.io.path.*
+
+@OptIn(ExperimentalPathApi::class)
+fun <T> usingTempDirectory(action: (path: Path) -> T): T {
```
dkrasnoff, line 11: nit: lets make it 4 spaces indents

gradle/libs.versions.toml
```diff
@@ -36,9 +38,10 @@ kotlin-compiler-arguments-description = { group = "org.jetbrains.kotlin", name =
 kotlin-script-runtime = { group = "org.jetbrains.kotlin", name = "kotlin-script-runtime", version.ref = "kotlin" }
 kotlin-gradlePlugin = { group = "org.jetbrains.kotlin", name = "kotlin-gradle-plugin", version.ref = "kotlin" }
 kotlin-dom-api-compat = { group = "org.jetbrains.kotlin", name = "kotlin-dom-api-compat", version.ref = "kotlin" }
-kotlin-compose-compiler-plugin = { group = "org.jetbrains.kotlin", name = "kotlin-compose-compiler-plugin", version.ref = "kotlin" }
+kotlin-compose-compiler-plugin = { group = "org.jetbrains.kotlin", name = "kotlin-compose-compiler-plugin-embeddable", version.ref = "kotlin" }
 kotlinx-coroutines-core-jvm = { group = "org.jetbrains.kotlinx", name = "kotlinx-coroutines-core-jvm", version.ref = "kotlinx-coroutines" }
 kotlinx-coroutines-reactor = { group = "org.jetbrains.kotlinx", name = "kotlinx-coroutines-reactor", version.ref = "kotlinx-coroutines" }
+kotlinx-coroutines-core-compose-wasm = { group = "org.jetbrains.kotlinx", name = "kotlinx-coroutines-core-wasm-js", version.ref = "kotlinx-coroutines-compose-wasm" }
```
dkrasnoff, line 46: Does it work for you? 
We have problems with coroutines bigger than 1.7.3 because of `executor.policy`

gradle/libs.versions.toml
```diff
@@ -7,18 +7,20 @@ aws-serverless-java-container-springboot3 = "2.1.5"
...
 lsp4j = "0.24.0"
 reactor = "3.8.0"
-skiko = "0.8.15"
+skiko = "0.9.22.2"
 testcontainers = "1.21.3"
 # don't forget to update jackson version in `executor.policy` file.
 jackson = "2.19.2"
 hamcrest = "3.0"
-compose = "1.7.0"
+compose = "1.9.3"
```
dkrasnoff, line 22: I recall, you told me not to dependabot update skiko, what about these to dependencies?

src/main/kotlin/com/compiler/server/utils/DependenciesUtil.kt
```diff
@@ -0,0 +1,22 @@
...
+import com.compiler.server.model.bean.VersionInfo
+import org.jetbrains.kotlin.arguments.dsl.base.KotlinCompilerArgument
+import org.jetbrains.kotlin.arguments.dsl.base.KotlinCompilerArguments
+import org.jetbrains.kotlin.arguments.dsl.base.KotlinCompilerArgumentsLevel
+import org.jetbrains.kotlin.arguments.dsl.types.*
+import org.jetbrains.kotlin.tooling.core.KotlinToolingVersion
+import org.springframework.stereotype.Component
+
+@Component
+class DependenciesUtil(
```
dkrasnoff: why do we need this wrapper if we can just call `kotlinEnvironment...` in place?

src/main/kotlin/com/compiler/server/service/KotlinProjectExecutor.kt
```diff
@@ -50,11 +71,15 @@ class KotlinProjectExecutor(
                     project,
                 ).compilerDiagnostics
 
-            ProjectType.WASM, ProjectType.COMPOSE_WASM ->
-                convertToWasm(
-                    project,
-                    debugInfo = false,
-                ).compilerDiagnostics
+            ProjectType.WASM -> convertToWasm(
```
dkrasnoff: nit: `ProjectType.WASM` and `ProjectType.COMPOSE_WASM` are identical, why do we need to separate them?

src/main/kotlin/com/compiler/server/model/bean/Dependencies.kt
```diff
@@ -0,0 +1,6 @@
+package com.compiler.server.model.bean
+
+data class Dependencies(
```
dkrasnoff: strange bean, why just don't use `DependenciesProperties` where needed?

src/main/kotlin/com/compiler/server/controllers/ResourceRestController.kt
```diff
@@ -1,64 +0,0 @@
...
-import org.springframework.http.*
-import org.springframework.web.bind.annotation.GetMapping
-import org.springframework.web.bind.annotation.RequestMapping
-import org.springframework.web.bind.annotation.RestController
-import java.util.concurrent.TimeUnit
-
-
-@RestController
-@RequestMapping(value = ["/api/resource", "/api/**/resource"])
-class ResourceRestController {
```
dkrasnoff, line 15: Should we change anything in our deployment process if this controller was removed?
cc @zoobestik

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
```diff
@@ -140,17 +143,20 @@ class KotlinToJSTranslator(
...
+                        WASM_DEFAULT_MODULE_NAME,
                     )
 
-                    ProjectType.COMPOSE_WASM -> Triple(
+                    ProjectType.COMPOSE_WASM -> WasmArguments(
                         composeWasmCompilerArguments,
                         compilerArgumentsUtil.PREDEFINED_COMPOSE_WASM_FIRST_PHASE_ARGUMENTS,
-                        compilerArgumentsUtil.PREDEFINED_COMPOSE_WASM_SECOND_PHASE_ARGUMENTS
+                        compilerArgumentsUtil.PREDEFINED_COMPOSE_WASM_SECOND_PHASE_ARGUMENTS +
+                                ("Xwasm-included-module-only" to true),
```
dkrasnoff: `Xwasm-included-module-only` is provided in compiler arguments, can we just add it to `PREDEFINED_COMPOSE_WASM_SECOND_PHASE_ARGUMENTS`?``

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
```diff
@@ -66,9 +58,20 @@ class KotlinToJSTranslator(
...
+        translate: (
+            List<ProjectFile>,
+            ProjectType,
+            Boolean,
+            JsCompilerArguments
+        ) -> CompilationResult<WasmTranslationSuccessfulOutput>
     ): TranslationResultWithJsCode {
         return try {
+            val outputFileName = when (projectType) {
+                ProjectType.WASM -> WASM_DEFAULT_MODULE_NAME
```
dkrasnoff, line 72: why should we have different `outputFileName` for wasm and compose wasm?

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
```diff
@@ -223,6 +254,11 @@ class KotlinToJSTranslator(
                         "  })();"
             ) + "\n export const instantiate = () => Promise.resolve();"
     }
+
+    private fun String.fixImports(): String = replace(
+        ".mjs",
+        "-${dependenciesUtil.dependenciesComposeWasm}.mjs"
```
dkrasnoff: not clear for me, why do we need such replace?

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
```diff
@@ -183,36 +189,61 @@ class KotlinToJSTranslator(
...
         val jsBuitinsAlias = JS_BUILTINS_ALIAS_NAME_REGEX.find(importObjectJsContent)?.groupValues?.get(1)
 
-        val replacedImportObjectContent = importObjectJsContent.replace(
-            JS_BUILTINS_ALIAS_NAME_REGEX,
-            "const $jsBuitinsAlias = await import(`data:application/javascript;base64, ${Base64.encode(wasmOutput.jsBuiltins.toByteArray())}`)"
-        )
+        val replacedImportObjectContent =
+            (wasmOutput.jsBuiltins?.toByteArray()?.let { byteContent ->
+                importObjectJsContent
+                    .replace(
```
dkrasnoff: nit: a little bit overwhelmed, replace, is it possible to extract some logic to a separate method, or just divide this logic to make it more readable?

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
```diff
@@ -240,7 +276,7 @@ private fun String.withMainArgumentsIr(arguments: List<String>): String {
 
 data class WasmTranslationSuccessfulOutput(
     val jsCode: String,
-    val jsBuiltins: String,
+    val jsBuiltins: String?,
```
dkrasnoff, line 298: In which case, it can be null?

resource-server/build.gradle.kts
```diff
@@ -0,0 +1,72 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
```
dkrasnoff, line 1: Could you add a README file to this module and mention there that it is needed only for testing, since it isn't obvious?

settings.gradle.kts
```diff
@@ -11,4 +11,6 @@ plugins {
 include(":executors")
 include(":common")
 include(":dependencies")
-include(":completions")
\ No newline at end of file
+include(":completions")
+include(":cache-maker")
```
dkrasnoff, line 15: It would be greate to add README to this module as well, because it is not obvious about which cache we are talking

build.gradle.kts
```diff
@@ -50,45 +66,77 @@ dependencies {
...
+    val hashValue: Provider<String> = kotlinComposeWasmRuntimeHash.elements.map {
+        it.single().asFile.readText()
     }
+
+    put(
+        dependenciesComposeWasm,
+        hashValue
+    )
+
+    put(
```
dkrasnoff, line 117: Now the property will look like:
```
dependencies.static-url=http://localhost:8081
```
I'd expect it to look like:
```
dependencies.static-url=${STATCI_URL:http://localhost:8081}
```

build.gradle.kts
```diff
@@ -119,15 +168,12 @@ val buildLambda by tasks.creating(Zip::class) {
...
 }
 
 tasks.named<Copy>("processResources") {
-    val archiveOperation = project.serviceOf<ArchiveOperations>()
-    from(resourceDependency.map {
-        archiveOperation.zipTree(it)
-    }) {
-        into("com/compiler/server")
-    }
+    dependsOn(propertiesGenerator)
```
dkrasnoff, line 179: I don't understand how this magic is working, but do we need `processResources` at all?

build.gradle.kts
```diff
@@ -119,15 +168,12 @@ val buildLambda by tasks.creating(Zip::class) {
     into("lib") {
         from(configurations.productionRuntimeClasspath) { exclude("tomcat-embed-*") }
     }
+
+    dependsOn(prepareComposeWasmResources)
```
dkrasnoff, line 175: For me, the content of the lambda zip is identical with and without this line. What should this line add to the lamda archive?

dkrasnoff (changes requested):

common/src/main/kotlin/com/compiler/server/common/components/CliUtils.kt
```diff
@@ -0,0 +1,25 @@
+package com.compiler.server.common.components
```
dkrasnoff, line 1: How often is it expected to change this build script?

cache-maker/src/wasmJsMain/kotlin/main.kt
ilgonmic: It is necessary to prepare caches for compose application. File is the same as the default compose sample which is used on kotlin-play

gradle/libs.versions.toml
ilgonmic: It works, because in case of coroutines for jvm, it is executed on server-side. But for wasm, it is executed locally on the user's machine, so `executor.policy` doesn't matter

gradle/libs.versions.toml
ilgonmic: The main idea - skiko should be updated with compose version. I don't know should we update compose version automatically.
The idea was that we can upgrade compose version manually, because for Compose we need to support only the latest version of Kotlin + some of (but not very old) version of Compose

gradle/libs.versions.toml
ilgonmic: Will add a comment

gradle/libs.versions.toml
ilgonmic: https://github.com/JetBrains/kotlin-compiler-server/blob/28bab182526ed54561934d27eecc20bd0145553b/.github/dependabot.yml

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
ilgonmic: Will add a comment

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
ilgonmic: Will divide it to separate methods

build.gradle.kts
ilgonmic: Now we keep it originally, then maybe we need to update later when we will deploy it

build.gradle.kts
ilgonmic: Will add a comment

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
ilgonmic: Renamed to `dependenciesComposeWasmHash`

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
ilgonmic: Compiler could generate this file if there are String usages (or other usages of builtins types), but this file is generated not in all cases

build.gradle.kts
ilgonmic: Done

build.gradle.kts
ilgonmic: Fixed, and done via tasks.processResources

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
ilgonmic: Done

src/main/kotlin/com/compiler/server/compiler/components/KotlinToJSTranslator.kt
ilgonmic: Done

src/main/kotlin/com/compiler/server/model/bean/Dependencies.kt
ilgonmic: Removed

src/main/kotlin/com/compiler/server/service/KotlinProjectExecutor.kt
ilgonmic: Done

src/main/kotlin/com/compiler/server/utils/DependenciesUtil.kt
ilgonmic: Done

gradle/libs.versions.toml
ilgonmic: Done

gradle/libs.versions.toml
ilgonmic: Done

common/src/main/kotlin/com/compiler/server/common/components/CliUtils.kt
ilgonmic: Done

cache-maker/src/wasmJsMain/kotlin/main.kt
ilgonmic: Done with Hello world application

settings.gradle.kts
dkrasnoff: Thanks a lot for the README files

PR #935, opened 2026-04-09
KTL-1731: Added valkey access auth and removed executor policy

nikitavlaevjb:


build.gradle.kts
```diff
@@ -99,10 +97,11 @@ fun Project.generateProperties(
...
     "wasm.compose.cache.enabled" to "\${CACHE_NAMESPACE:false}",
     "spring.data.redis.host" to "=\${SPRING_DATA_REDIS_HOST:}",
-    "spring.data.redis.port" to "\${SPRING_DATA_REDIS_SSL_ENABLED:6379}",
-    "speing.data.redis.ssl.enabled" to "\${SPRING_DATA_REDIS_SSL_ENABLED:false}",
+    "spring.data.redis.port" to "\${SPRING_DATA_REDIS_PORT:6379}",
+    "spring.data.redis.ssl.enabled" to "\${SPRING_DATA_REDIS_SSL_ENABLED:false}",
     "spring.data.redis.connectTimeout" to "1000",
-    "spring.data.redis.timeout" to "1000"
+    "spring.data.redis.timeout" to "1000",
+    "valkey.secret.arn" to "\${VALKEY_SECRET_ARN:}"
```
dkrasnoff: I wonder will it be possible to then read `application.yaml` somehow from the user's side?

src/main/kotlin/com/compiler/server/executor/JavaExecutor.kt
```diff
@@ -17,10 +16,13 @@ class JavaExecutor {
   companion object {
     const val MAX_OUTPUT_SIZE = 100 * 1024
     const val EXECUTION_TIMEOUT = 10000L
+
   }
 
   fun execute(args: List<String>): ProgramOutput {
-    return Runtime.getRuntime().exec(args.toTypedArray()).use {
+    val processBuilder = ProcessBuilder(args)
```
dkrasnoff, line 22: I suspect that this is a cause of test failures.

src/main/kotlin/com/compiler/server/configuration/RedisConfiguration.kt
```diff
@@ -0,0 +1,48 @@
...
+
+        val clientConfig = LettuceClientConfiguration.builder()
+        if (sslEnabled) {
+            clientConfig.useSsl()
+        }
+
+        return LettuceConnectionFactory(config, clientConfig.build())
+    }
+
+    private fun loadPasswordFromSecretsManager(arn: String): String {
```
dkrasnoff: Not clear to me what prevents the user from calling the same code and getting the same secrets?

dkrasnoff (changes requested):

build.gradle.kts
```diff
@@ -99,10 +97,11 @@ fun Project.generateProperties(
     "springdoc.swagger-ui.path" to "/api-docs/swagger-ui.html",
     "wasm.compose.cache.enabled" to "\${CACHE_NAMESPACE:false}",
     "spring.data.redis.host" to "=\${SPRING_DATA_REDIS_HOST:}",
-    "spring.data.redis.port" to "\${SPRING_DATA_REDIS_SSL_ENABLED:6379}",
-    "speing.data.redis.ssl.enabled" to "\${SPRING_DATA_REDIS_SSL_ENABLED:false}",
+    "spring.data.redis.port" to "\${SPRING_DATA_REDIS_PORT:6379}",
```
dkrasnoff: just wonder, how does it work in production now

dkrasnoff: Hm, maybe if it cannot read the port (in our case, it is true), it uses the default port

src/main/kotlin/com/compiler/server/configuration/RedisConfiguration.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.data.redis.connection.RedisStandaloneConfiguration
+import org.springframework.data.redis.connection.lettuce.LettuceClientConfiguration
+import org.springframework.data.redis.connection.lettuce.LettuceConnectionFactory
+import software.amazon.awssdk.services.secretsmanager.SecretsManagerClient
+import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueRequest
+
+@Configuration
+@ConditionalOnProperty("wasm.compose.cache.enabled")
+class RedisConfiguration(
+    @Value("\${spring.data.redis.host}") private val host: String,
```
dkrasnoff: We can just use an environment variable here, as I understand, instead of using application.properties.

src/main/kotlin/com/compiler/server/compiler/components/KotlinCompiler.kt
```diff
@@ -68,7 +65,7 @@ class KotlinCompiler(
             ?.joinToString("\n\n")
     }
 
-    fun run(
+    fun  run(
```
dkrasnoff: extra space

build.gradle.kts
nikitavlaevjb: Well, arn is not dangerous to leak at all.
Whats dangerous is such path:
```
  fun main() {
      // Step 1: Read parent process environment (same UID, so /proc is readable)
      val ppid = ProcessHandle.current().parent().get().pid()
      val environ = java.io.File("/proc/$ppid/environ").readText()
      // This exposes: AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY,
      //               AWS_SESSION_TOKEN, VALKEY_SECRET_ARN,
      //               SPRING_DATA_REDIS_HOST, etc.

      // Step 2: Use the AWS creds to call Secrets Manager VPC endpoint
      //         → retrieve actual Valkey auth token

      // Step 3: Connect to Valkey on port 6379 (same VPC, allowed by SG)
      //         → read/write/delete cached data

      // Step 4: Print everything to stdout → returned in API response
      println(environ)
  }
```
I'm working on it

cache-proxy/src/main/kotlin/com/compiler/server/cacheproxy/CacheProxyHandler.kt
```diff
@@ -0,0 +1,190 @@
...
+    override fun handleRequest(input: java.io.InputStream, output: java.io.OutputStream, context: Context) {
+        val log = context.logger
+        val event = mapper.readTree(input)
+        val path = event.get("path")?.asText() ?: ""
+        val body = event.get("body")?.asText()
+        val queryParams = event.get("queryStringParameters")
+        val compiler = queryParams?.get("compiler")?.asText()
+        log.log("cache-proxy: path=$path compiler=$compiler bodyLength=${body?.length ?: 0}\n")
+
+        val cacheKey = when {
```
dkrasnoff: Could you please move the cache logic to a separate service, as it was made before?

cache-proxy/build.gradle.kts
```diff
@@ -0,0 +1,24 @@
+plugins {
+    id("base-kotlin-jvm-conventions")
+}
+
+dependencies {
+    implementation(libs.aws.lambda.core)
+    implementation(libs.aws.lambda.sdk)
+    implementation(libs.lettuce.core)
+    implementation("org.slf4j:slf4j-simple:1.7.36")
```
dkrasnoff: use libs.versions.toml

gradle/libs.versions.toml
```diff
@@ -3,7 +3,10 @@ kotlin = "2.4.0-Beta1"
 spring-boot = "3.5.7"
 spring-dependency-managment = "1.1.7"
 springdoc = "2.8.14"
+aws-sdk = "2.29.0"
 aws-serverless-java-container-springboot3 = "2.1.5"
+aws-lambda-java = "1.2.3"
+lettuce = "6.5.5.RELEASE"
```
dkrasnoff: IDEA shows me that this version contains vulnarability, lets use newer version if it is a new library anyway

cache-proxy/src/main/kotlin/com/compiler/server/cacheproxy/dto/ComposeWasmRequest.kt
```diff
@@ -0,0 +1,21 @@
+package com.compiler.server.cacheproxy.dto
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+
+@JsonIgnoreProperties(ignoreUnknown = true)
```
dkrasnoff: wdyt about sharing these dto's between main and this module, for example via common module?
In such a case we will not fall into the contract differens, when something was changed in the compiler module and was not changed here.

cache-proxy/build.gradle.kts
dkrasnoff, line 1: Add a README.md file for this module, please. 
Describe there:
* the purpose of the module
* small description of the architecture (it differs from the main module)
* how to build it + how to run locally

cache-proxy/src/test/kotlin/com/compiler/server/cacheproxy/CacheKeyCompatibilityTest.kt
```diff
@@ -0,0 +1,150 @@
...
+                FileDto(text = "class B", name = "B.kt"),
+            ),
+        )
+        val request2 = ProjectRequest(
+            files = listOf(
+                FileDto(text = "class B", name = "B.kt"),
+                FileDto(text = "class A", name = "A.kt"),
+            ),
+        )
+        assertEquals(buildV1CacheKey(request1), buildV1CacheKey(request2))
```
dkrasnoff: Very strange test and check, it checks the `buildV1CacheKey` and `buildV1CacheKey`, which are literally declared in this test class. 
So, when you change the logic of creating the key in the source code, the test will still pass...
This comment is actual for all the tests in this class, I asume

cache-proxy/src/main/kotlin/com/compiler/server/cacheproxy/CacheProxyHandler.kt
```diff
@@ -0,0 +1,190 @@
...
+        .configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true)
+        .build()
+
+    private val targetLambda = System.getenv("TARGET_LAMBDA_NAME")
+    private val kotlinVersion = System.getenv("KOTLIN_VERSION")
+    private val cacheNamespace = System.getenv("CACHE_NAMESPACE") ?: "default"
+    private val ttlSeconds = 86400L // 24 hours
+
+    private val lambdaClient: LambdaClient = LambdaClient.create()
+    private val redis: RedisCommands<String, String> = createRedisConnection()
```
dkrasnoff: I'm not sure it's okay to have a single connection for the whole session.

cache-proxy/src/main/kotlin/com/compiler/server/cacheproxy/CacheProxyHandler.kt
```diff
@@ -0,0 +1,190 @@
...
+                try {
+                    val request = mapper.readValue<ComposeWasmV2Request>(body)
+                    buildV2CacheKey(request)
+                } catch (e: Exception) {
+                    log.log("cache-proxy: V2 parse failed: ${e.message}\n")
+                    null
+                }
+            }
+            // V1: /translate?compiler=compose-wasm
+            compiler == "compose-wasm" && body != null -> {
```
dkrasnoff: maybe then it is better to check combination `/translate` and `compose-wasm` arg

cache-proxy/src/main/kotlin/com/compiler/server/cacheproxy/CacheProxyHandler.kt
```diff
@@ -0,0 +1,190 @@
...
+        val hash = MessageDigest.getInstance("SHA-256")
+            .digest(idempotentMapper.writeValueAsString(normalized).toByteArray())
+            .joinToString("") { "%02x".format(it) }
+        return "$prefix:$cacheNamespace:v$kotlinVersion:$hash"
+    }
+
+    private fun hasErrors(body: String): Boolean {
+        val node = mapper.readTree(body)
+        if (node.get("exception")?.isNull == false) return true
+        node.get("errors")?.fields()?.forEach { (_, diagnostics) ->
```
dkrasnoff: "This method is deprecated, use properties() instead."

cache-proxy/src/main/kotlin/com/compiler/server/cacheproxy/CacheProxyHandler.kt
```diff
@@ -0,0 +1,190 @@
...
+
+    private val targetLambda = System.getenv("TARGET_LAMBDA_NAME")
+    private val kotlinVersion = System.getenv("KOTLIN_VERSION")
+    private val cacheNamespace = System.getenv("CACHE_NAMESPACE") ?: "default"
+    private val ttlSeconds = 86400L // 24 hours
+
+    private val lambdaClient: LambdaClient = LambdaClient.create()
+    private val redis: RedisCommands<String, String> = createRedisConnection()
+
+    override fun handleRequest(input: java.io.InputStream, output: java.io.OutputStream, context: Context) {
```
dkrasnoff: I suggest dividing this method into smaller parts, even creating not only a service for caching, but also a service for request processing or forwarding.

dkrasnoff (changes requested):

src/test/kotlin/com/compiler/server/WasmComposeCacheServiceIntegrationTest.kt
dkrasnoff, line 1: It would be great to reproduce the whole check in a new module

cache-proxy/src/main/kotlin/com/compiler/server/cacheproxy/dto/ComposeWasmRequest.kt
nikitavlaevjb: Makes sense

cache-proxy/src/main/kotlin/com/compiler/server/cacheproxy/CacheProxyHandler.kt
```diff
@@ -0,0 +1,85 @@
...
+            "body" to body,
+            "isBase64Encoded" to false,
+        )
+    )
+
+    private fun extractBodyToCache(response: String, log: LambdaLogger): String? {
+        val node = mapper.readTree(response)
+        val statusCode = node.get("statusCode")?.asInt() ?: 500
+        val body = node.get("body")?.asText()
+        if (statusCode != 200 || body == null) {
```
dkrasnoff: nit: I see that there is `HttpStatusCode.OK` and `HttpStatusCode.INTERNAL_SERVER_ERROR` for 200 and 500

cache-proxy/src/main/kotlin/com/compiler/server/cacheproxy/CacheProxyHandler.kt
```diff
@@ -0,0 +1,85 @@
...
+import software.amazon.awssdk.services.lambda.LambdaClient
+import java.io.InputStream
+import java.io.OutputStream
+import java.time.Duration
+
+// @JvmOverloads generates a no-arg constructor using default parameter values, which is required for Lambda
+class CacheProxyHandler @JvmOverloads constructor(
+    private val cacheService: CacheService = CacheService(
+        redis = createRedisConnectionFromEnv(),
+        kotlinVersion = System.getenv("KOTLIN_VERSION"),
```
dkrasnoff: I thought that it is expected to have only one cache proxy lambda, which will process all the versions. 
Why do we hardcode KOTLIN_VERSION via environment variables rather than getting it from the request URL?

cache-proxy/src/main/kotlin/com/compiler/server/cacheproxy/service/CacheEndpointType.kt
```diff
@@ -1,6 +1,6 @@
-package com.compiler.server.enums
+package com.compiler.server.cacheproxy.service
```
dkrasnoff: com.compiler.server.cacheproxy.enums

cache-proxy/src/main/kotlin/com/compiler/server/cacheproxy/service/LambdaForwarder.kt
```diff
@@ -0,0 +1,29 @@
+package com.compiler.server.cacheproxy.service
```
dkrasnoff: com.compiler.server.cacheproxy.utils

cache-proxy/src/main/kotlin/com/compiler/server/cacheproxy/service/RedisConnection.kt
```diff
@@ -0,0 +1,22 @@
+package com.compiler.server.cacheproxy.service
```
dkrasnoff: com.compiler.server.cacheproxy.utils or com.compiler.server.cacheproxy.configuration

cache-proxy/src/main/kotlin/com/compiler/server/cacheproxy/service/RequestParsing.kt
```diff
@@ -0,0 +1,29 @@
+package com.compiler.server.cacheproxy.service
```
dkrasnoff: com.compiler.server.cacheproxy.utils

dkrasnoff (changes requested):

dkrasnoff (approved):

src/main/kotlin/com/compiler/server/controllers/CompilerRestController.kt
```diff
@@ -84,10 +86,18 @@ class CompilerRestController(
         )
     }
 
+    // TranslateComposeWasmRequest lives in :common (so the cache-proxy can deserialize it)
```
dkrasnoff: I don't think that we really need this comment

dkrasnoff (approved):

src/main/kotlin/com/compiler/server/controllers/CompilerRestController.kt
nikitavlaevjb: fixed

PR #969, opened 2026-07-07
KTL-4631: reproduced JavaExecutor native-thread leak and fixed pool leak

nikitavlaevjb:
TODO: Probably need a better place for this e2e test (if any).
It takes ~30s.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -194,22 +194,9 @@ tasks.withType<Test> {
     }
 }
 
-// TODO move this somewhere appropriate
-// KTL-4631: heavy container-based reproduction of the JavaExecutor native-thread leak.
-// Excluded from the default `test` task; run explicitly via `./gradlew leakTest`.
 tasks.named<Test>("test") {
-    exclude("**/ThreadLeakE2ETest*")
-}
-
-val leakTest by tasks.registering(Test::class) {
-    description = "KTL-4631 e2e: reproduce the JavaExecutor native-thread leak in a pid-capped container."
-    group = "verification"
-    testClassesDirs = sourceSets["test"].output.classesDirs
-    classpath = sourceSets["test"].runtimeClasspath
-    include("**/ThreadLeakE2ETest*")
-    systemProperty("e2e.kotlin.version", libs.versions.kotlin.get())
+    // ThreadLeakE2ETest needs the Kotlin version to locate the boot jar
+    systemProperty("kotlin.version", libs.versions.kotlin.get())
     // Slim image copies the host-built boot jar, so build it first.
     dependsOn(tasks.named("bootJar"))
-    // Always re-run: the outcome depends on the live container, not on inputs.
-    outputs.upToDateWhen { false }
 }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -72,6 +72,7 @@ dependencies {
     testImplementation(libs.kotlin.test)
     testImplementation(libs.kotlinx.coroutines.test)
     testImplementation(libs.kotlin.mockito)
+    testImplementation(libs.bundles.testcontainers)
 
     kotlinComposeWasmCache(project(":cache-maker"))
 }
@@ -191,4 +192,24 @@ tasks.withType<Test> {
         this@withType.environment("kotlin.wasm.node.path", executablePath.get())
         this@withType.environment("kotlin.compose.wasm.resources.path", composeWasmResourcesPath.get())
     }
+}
+
+// TODO move this somewhere appropriate
+// KTL-4631: heavy container-based reproduction of the JavaExecutor native-thread leak.
+// Excluded from the default `test` task; run explicitly via `./gradlew leakTest`.
+tasks.named<Test>("test") {
+    exclude("**/ThreadLeakE2ETest*")
+}
+
+val leakTest by tasks.registering(Test::class) {
+    description = "KTL-4631 e2e: reproduce the JavaExecutor native-thread leak in a pid-capped container."
+    group = "verification"
+    testClassesDirs = sourceSets["test"].output.classesDirs
+    classpath = sourceSets["test"].runtimeClasspath
+    include("**/ThreadLeakE2ETest*")
+    systemProperty("e2e.kotlin.version", libs.versions.kotlin.get())
+    // Slim image copies the host-built boot jar, so build it first.
+    dependsOn(tasks.named("bootJar"))
+    // Always re-run: the outcome depends on the live container, not on inputs.
+    outputs.upToDateWhen { false }
 }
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/compiler/server/executor/JavaExecutor.kt
```diff
@@ -30,9 +30,13 @@ class JavaExecutor {
       val standardOutput = asyncBufferedOutput(standardOut, limit = MAX_OUTPUT_SIZE)
       val errorOutput = asyncBufferedOutput(standardError, limit = MAX_OUTPUT_SIZE)
 
+      // one thread per output; must be shut down after use, otherwise its non-daemon
+      // worker threads linger and leak (KTL-4631: "unable to create native thread").
+      val outputPool = Executors.newFixedThreadPool(2)
+
       try {
         val currTime = System.currentTimeMillis()
-        val futuresList = Executors.newFixedThreadPool(2) // one thread per output
```
dkrasnoff, line 35: Interesting catch 👍 
I see that this code has been with us since 2019. Why didn't we catch this problem before? Have you tried to reproduce it on older branches?

src/test/kotlin/com/compiler/server/ThreadLeakE2ETest.kt
```diff
@@ -0,0 +1,136 @@
...
+        .withFileFromString("Dockerfile", dockerfile)
+        .withFileFromPath("app.jar", jarPath)
+        .withFileFromPath(kotlinVersion, Paths.get(kotlinVersion).toAbsolutePath())
+
+    private val container: GenericContainer<*> = GenericContainer(image)
+        .withExposedPorts(PORT)
+        .withCreateContainerCmdModifier { cmd ->
+            cmd.hostConfig
+                ?.withPidsLimit(PIDS_LIMIT)
+                ?.withMemory(4L * 1024 * 1024 * 1024) // 4 GiB: comfortable heap so the leak surfaces as pid exhaustion, not heap OOM
```
dkrasnoff: nit: our lambdas have 2G, lets make it like this?

src/test/kotlin/com/compiler/server/ThreadLeakE2ETest.kt
```diff
@@ -0,0 +1,136 @@
...
+
+    @Test
+    fun `run endpoint does not leak native threads under repeated calls`() {
+        val client = HttpClient.newBuilder()
+            .connectTimeout(Duration.ofSeconds(REQUEST_TIMEOUT_SECONDS))
+            .build()
+
+        val baseUrl = "http://${container.host}:${container.getMappedPort(PORT)}"
+        val body = jacksonObjectMapper().writeValueAsString(
+            mapOf(
```
dkrasnoff: Let's use real request body just in case
```
            RunRequest(
                args = "",
                files = listOf(ProjectFileRequestDto(name = "File.kt", text = "fun main() { println(\"ok\") }")),
            )
```

src/test/kotlin/com/compiler/server/ThreadLeakE2ETest.kt
```diff
@@ -0,0 +1,136 @@
...
+                fail<Nothing>(
+                    "Native-thread leak reproduced at iteration $iteration/$MAX_ITERATIONS: " +
+                        "request failed with ${e.javaClass.simpleName}: ${e.message}"
+                )
+            }
+
+            val payload = response.body() ?: ""
+            val marker = NATIVE_THREAD_MARKERS.firstOrNull { payload.contains(it, ignoreCase = true) }
+
+            if (response.statusCode() >= 500 || marker != null) {
```
dkrasnoff: Now, if the test catches 400, for example, in the response body, it passes, so we can get into a situation where the test starts to pass, because we receive an error with another code.
Let's make it fail even if the response code and response body have changed.

build.gradle.kts
```diff
@@ -192,3 +193,23 @@ tasks.withType<Test> {
         this@withType.environment("kotlin.compose.wasm.resources.path", composeWasmResourcesPath.get())
     }
 }
+
+// TODO move this somewhere appropriate
+// KTL-4631: heavy container-based reproduction of the JavaExecutor native-thread leak.
+// Excluded from the default `test` task; run explicitly via `./gradlew leakTest`.
+tasks.named<Test>("test") {
+    exclude("**/ThreadLeakE2ETest*")
```
dkrasnoff: Let it be in all tests execution, I think it is important test to be included even if it runs around 1m

dkrasnoff: I mean lets just replace all with:
```
tasks.named<Test>("test") {
    systemProperty("kotlin.version", libs.versions.kotlin.get())
}
```

src/main/kotlin/com/compiler/server/executor/JavaExecutor.kt
nikitavlaevjb: checked out 2020 code, leak confirmed

build.gradle.kts
nikitavlaevjb: fixed

src/test/kotlin/com/compiler/server/ThreadLeakE2ETest.kt
nikitavlaevjb: fixed

src/test/kotlin/com/compiler/server/ThreadLeakE2ETest.kt
nikitavlaevjb: fixed

src/test/kotlin/com/compiler/server/ThreadLeakE2ETest.kt
nikitavlaevjb: fixed

src/main/kotlin/com/compiler/server/executor/JavaExecutor.kt
dkrasnoff: Thanks a lot

src/test/kotlin/com/compiler/server/ThreadLeakE2ETest.kt
dkrasnoff: nit: I actually still don't like the format with a generic message "Native-thread leak (or unexpected response) at iteration ".
I'd prefer to have two separate assertions, but you can leave as is

dkrasnoff:
Hey @nikitavlaevjb, this pull request should go via [hotfix flow](https://github.com/JetBrains/kotlin-compiler-server/blob/master/docs/workflow.md#hotfix-flow), and in your case, in master as well.

dkrasnoff (approved):

PR #975, opened 2026-07-10
KTL-4698: Handle internal compiler errors in `CompilationLogger`

dkrasnoff:


src/main/kotlin/com/compiler/server/compiler/components/CompilationLogger.kt
nikitavlaevjb, line 58: Should we update warn as well?

src/main/kotlin/com/compiler/server/compiler/components/CompilationLogger.kt
```diff
@@ -84,4 +86,32 @@ class CompilationLogger(
...
+                if (isNotEmpty()) append("\n")
+                append(throwable.stackTraceToString())
+            }
+        }
+        if (message.isEmpty()) return
+        val ed = ErrorDescriptor(
+            interval = null,
+            message = message,
+            severity = ProjectSeveriry.ERROR,
+            className = ProjectSeveriry.ERROR.name
```
nikitavlaevjb: Why we set className to "ERROR"?

src/main/kotlin/com/compiler/server/compiler/components/CompilationLogger.kt
```diff
@@ -84,4 +86,32 @@ class CompilationLogger(
...
+        )
+        compilationLogs.values.firstOrNull()?.add(ed)
```
nikitavlaevjb, line 115: Can we try to recover the actual file the error belongs to instead of taking the first?

nikitavlaevjb (changes requested):

src/main/kotlin/com/compiler/server/compiler/components/CompilationLogger.kt
dkrasnoff: Good question. I've checked that for the compilation errors we previously received:
```
{
    "errors": {
        "File.kt": [
            {
                "message": "<StackTrace>",
                "severity": "ERROR",
                "className": "ERROR"
            }
        ]
    },
    "exception": null,
    "jvmByteCode": null,
    "text": ""
}
```
And made the same here.

src/main/kotlin/com/compiler/server/compiler/components/CompilationLogger.kt
dkrasnoff: I expect that we get in `addFallbackCompilationLog` only during compilation errors, so I'm not sure that we can restore the actual file path. At the same time we now have only one file `File.kt`, so it is kind of workaround.

src/main/kotlin/com/compiler/server/compiler/components/CompilationLogger.kt
dkrasnoff: Checked, we shouldn't because there are compiler warnings, which we shouldn't show like "Following manually enabled features will force generation of pre-release binaries: ExplicitBackingFields"

nikitavlaevjb (approved):

